PR #155, opened 2026-04-15
feat(vize_armature): add CDATA support in tokenizer and parser

Jevon617:


ubugeeei:
Thank you.

I've handled on https://github.com/ubugeeei/vize/pull/157 with Co-Author

PR #208, opened 2026-04-27
fix: correct Vue prop source spans

### LINKED ISSUE #206: Wrong span returned by `vize_armature` for Vue props
The reproduction repo: https://github.com/liang-demos/vize-armature-directives-spaces

### Case One: Trailing whitespaces after directive names

For `<div v-if />`, we hope the `head_space` we got is just `v-if`'s location, but while we actually get `v-if ` (One whitespace there)

### Case Two: Incomplete full span

For `<div class ="w-100" />`, we expect to get the `full_span` just fit for `class ="w-100"`, but we actually get `class ="w-100` (One `"` missing). I guess it was caused by doing `+1` for skipping `=` directly without caring about whitespaces beside eq sign.

ubugeeei:
## Summary
- Correct parser prop source locations so no-value directives stop at the directive head instead of including trailing whitespace.
- Include the closing quote in quoted attribute and directive full locations, including spaced equals.
- Update diagnostics snapshots that use the corrected prop spans.

## Tests
- `cargo test -p vize_armature`
- `cargo test -p vize_patina`

Closes #206

### DIFF
--- a/crates/vize_armature/src/parser/attribute.rs
+++ b/crates/vize_armature/src/parser/attribute.rs
@@ -28,6 +28,17 @@ impl<'a> Parser<'a> {
         });
     }
 
+    /// Process the end of a full attribute or directive head.
+    pub(super) fn on_attrib_name_end_impl(&mut self, end: usize) {
+        if let Some(ref mut attr) = self.current_attr {
+            attr.name_end = end;
+        }
+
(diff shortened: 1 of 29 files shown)


liangmiQwQ:
Thank you

ubugeeei:
@liangmiQwQ 
Thank you for always keeping me updated!
Please take a look at these as well!

PR #285, opened 2026-05-15
fix(lsp): suppress type diagnostics after parse errors

ubugeeei:
## Summary
- stop LSP diagnostic collection after an SFC parser error so broken files do not also emit derived type/lint noise
- skip async Corsa diagnostics when sync collection already reported an SFC parser error
- add regressions for malformed SFC type checking and LSP diagnostic aggregation

## Verification
- cargo fmt --check --all
- git diff --check
- vp check --fix crates/vize_maestro/src/ide/diagnostics/mod.rs crates/vize_canon/src/sfc_typecheck/mod.rs

Focused cargo tests were deferred to CI because the local rustc is 1.92.0 while current OXC dependencies require 1.93.0.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -55,33 +55,55 @@ jobs:
         working-directory: head
         run: vp install --frozen-lockfile --prefer-offline
 
+      - name: Validate base metadata
+        id: validate-base
+        continue-on-error: true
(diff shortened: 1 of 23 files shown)


ubugeeei:
Superseded by a fresh branch so pull_request checks can run without force-pushing.

Issue #503, opened 2026-05-18
release: declare and enforce workspace MSRV (rust-version) in Cargo.toml


ubugeeei:
## Problem

`rust-toolchain.toml` pins the development toolchain to `1.95.0`, but the workspace `Cargo.toml` does not declare `rust-version`. Adopters who build from source against a different toolchain have no guarantee about which version is supported. Downstream packagers (Nix, Homebrew, Debian) usually require an explicit MSRV.

## Goal

- Add `rust-version = "1.x.y"` to `[workspace.package]` (and inherit it with `rust-version.workspace = true` in each crate).
- Set the MSRV to the oldest supported version, ideally not the same as the development pin (e.g. dev = 1.95, MSRV = 1.91).
- Add a CI job that runs `cargo +<MSRV> check --workspace` to enforce it.
- Document the MSRV policy in `docs/release/stability.md` (e.g. 'we support the latest stable and the two previous releases').

## Acceptance criteria

- [ ] MSRV declared and enforced in CI.
- [ ] Policy documented (cadence for bumping the MSRV, how it relates to releases).
- [ ] Pre-1.0 break for an MSRV bump is mentioned in the changelog template.

ubugeeei:
進捗 (実装ブランチで対応中):

- ✅ \`Cargo.toml\` の \`[workspace.package]\` に \`rust-version = "1.95.0"\` を宣言
- ✅ 全 19 crate (publish/internal/test 含む) が \`rust-version.workspace = true\` を継承
- ✅ \`cargo metadata\` 出力で 19/19 が \`rust-version=1.95.0\` を持つことを確認
- ✅ \`CONTRIBUTING.md\` に MSRV ポリシーを追記
- ✅ \`docs/content/stability.md\` の Runtime Support 節に MSRV ポリシー段落を追加

残タスク:

- [ ] MSRV を実際の最小値 (たとえば dev=1.95, MSRV=1.91) まで下げて \`cargo +<MSRV> check\` を CI で常時実行
  - 現状の値はトolchain pin と同一なため、policy 宣言以上の検証強度はない
  - edition 2024 を維持する以上 1.85 が下限。落とせる幅を別 PR で検証する

ubugeeei:
[...]

PR #504, opened 2026-05-18
chore(release): declare workspace MSRV (rust-version = 1.95.0)

ubugeeei:
## Summary

- Declare `[workspace.package].rust-version = "1.95.0"` in `Cargo.toml`.
- Propagate it via `rust-version.workspace = true` to all 19 crates (publishable, workspace-only, and the test runner).
- Add an MSRV note to the `CONTRIBUTING.md` Setup section.
- Add an MSRV policy paragraph to the `docs/content/stability.md` Runtime Support section.

Without an explicit MSRV, downstream packagers (Nix, Homebrew, Debian) and `cargo install` users could not reason about which toolchain Vize promises to support. This PR makes the contract explicit.

## Why 1.95.0

The initial MSRV matches the development toolchain currently pinned in `rust-toolchain.toml`. As long as edition 2024 is in use, the absolute floor is 1.85.0, but verifying how far the MSRV can actually move down is left for a follow-up PR that adds CI enforcement against an older toolchain.

## Verification

```sh
cargo metadata --no-deps --format-version 1 --offline \
  | python3 -c "import json,sys; [print(p['name'], p.get('rust_version')) for p in json.load(sys.stdin)['packages']]"
```

- 19/19 crates report `rust-version = 1.95.0`.

## Test plan

- [ ] CI `Check` workflow is green.
- [ ] `cargo metadata` shows `rust-version` on every crate.
- [ ] Downstream packagers can read `rust-version` from any crate's `Cargo.toml`.

## Related

Closes part of #503. CI enforcement of the declared MSRV (running `cargo +<MSRV> check`) is left for a follow-up.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,7 +4,9 @@ Thanks for helping make Vize sharper. This project is still moving toward v1 alp
 
 ## Setup
 
-Use the Node.js version from `.node-version`, then install dependencies from the workspace root:
+Use the Node.js version from `.node-version` and the Rust version from `rust-toolchain.toml`. The workspace declares a minimum supported Rust version (MSRV) of `1.95.0` in `Cargo.toml` (`[workspace.package].rust-version`); contributions must compile under that version.
+
+Install dependencies from the workspace root:
 
 ```sh
 vp install --frozen-lockfile --prefer-offline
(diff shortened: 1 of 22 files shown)


ubugeeei:
**CI note**: `app-e2e (check)` failure is unrelated to this PR — it's a pre-existing snapshot mismatch tracked in #510 (added in `dd7e803b` and red on every PR since). `Fresh install smoke` failures should clear once #506 lands on `main` and this branch rebases.

PR #761, opened 2026-05-30
fix(croquis): narrow discriminated unions in flat v-else / v-else-if branches (#5850)

ubugeeei:
## What

Fixes a type-check parity divergence between Vize's editor layer and `vue-tsc` 6 / Vue 3.6, driven by `vuejs/language-tools#5850`, verified end-to-end against real `vue-tsc` through Corsa.

When the parser keeps a `v-if` chain as sibling elements (not grouped into an `IfNode`), the guard was built only from the element's own condition — preceding sibling conditions were never negated and `v-else` got no guard, so `v-else` branches type-checked **unnarrowed** (false **TS2339** when accessing the other union variant). Now a branch chain is tracked: `v-if` opens it, `v-else-if`/`v-else` negate every preceding condition, and a non-conditional sibling resets it.

## Tests
- croquis: `flat_v_else_branch_gets_negated_guard`, `flat_v_else_if_chain_accumulates_negated_guards`, `non_conditional_sibling_breaks_v_if_chain`
- check_cli (end-to-end via Corsa): `check_v_else_branch_narrows_discriminated_union`
- canon/croquis insta snapshots and the `node --test` check snapshots refresh only the `virtualTs` guard representation — diagnostic counts/messages are unchanged.

## Verification
`cargo test -p vize_croquis` (230), `cargo test -p vize_canon` (167), check_cli green; compiler oracle 625/625; fmt/clippy clean.

> The originally-bundled `v-for` key/value typing fix (#5978) was **split out** — its `__vForList` virtual-TS helper regressed the `Type check` benchmark by +11.6% (over the 5% budget), so it needs perf work before landing. Tracked in the follow-up issue.


ubugeeei:
Rebased onto current `main` (which now includes #759/#760/#762) and refreshed the check snapshots. All checks pass **except `cargo-semver-checks`**, which is red because `main` inherited the intentional public-API break from #762 (added fields to `vize_atelier_sfc`'s published structs). Per the decision to accept that break, it resolves at the next release via a crate version bump — tracked in #768. This PR's own changes (croquis v-else narrowing) introduce no API break.

PR #997, opened 2026-06-04
fix(sfc): emit user imports exactly once when both script blocks are present

### LINKED ISSUE #993: fix: side-effect imports run twice when SFC has both <script> and <script setup>
## Problem

When an SFC has both a normal `<script>` block and a `<script setup>` block, vize's inline-script preamble emits the user's imports twice. Any top-level side effect in those imports — registering a hotkey listener, pushing into a global registry, calling `installPlugin()` — runs twice in the final bundle.

Code-grounded finding in
`crates/vize_atelier_sfc/src/compile_script/inline/compiler/preamble.rs`:

- L118-120: imports from the preserved normal `<script>` are extracted via `parse_script_content(...)`.
- L126-128: \`deduped_imports\` (the imports from `<script setup>`) are written to \`output\`. Any imports that also live in the normal `<script>` are NOT deduped against \`normal_script_imports\` here — they're collected into \`setup_return_imports\` for the template-binding map only.
- L150: the entire \`preserved_normal_script\` body — which still contains its own \`import\` statements verbatim — is appended to \`output\` after the TypeScript declarations.

So a fixture like

```vue
<script lang=\"ts\">
import { registerHotkey } from './hotkey-plugin'
registerHotkey()
</script>

<script setup lang=\"ts\">
import { ref } from 'vue'
const count = ref(0)
</script>

<template>{{ count }}</template>
```

emits both `import { registerHotkey } from './hotkey-plugin'` AND the same import again inside the appended normal-script block. The bundler resolves each ESM specifier once, but the module's top-level statements execute once per import statement in the SAME compiled module, so \`registerHotkey()\` fires twice. The vite-plugin path (\`npm/vite-plugin-vize/src/plugin/load.ts:144-214\`) requests the SFC once — the duplication is purely in the SFC compile output.

## Expected outcome

The preamble emits each user import exactly once. The normal `<script>` block's body still flows through, but its `import` statements are stripped before emission (they're already in `normal_script_imports`, which is already covered by the deduped emit path).

## Acceptance criteria

- A regression fixture in `crates/vize_atelier_sfc/src/compile/snapshots` covers an SFC with both `<script>` and `<script setup>` importing from the same module — the compiled output contains exactly one `import` statement for that specifier.
- A separate fixture with both blocks importing **different** modules still emits both, but each exactly once.
- The existing `compile_both_script_blocks` snapshot is refreshed to drop the duplicate \`import\` line.
- Behavior parity check: a fixture whose normal `<script>` runs `console.log("once")` at module scope produces output that, when evaluated, prints `"once"` exactly once.

mvanhorn:
## Summary

When an SFC has both a normal `<script>` and a `<script setup>`, the inline preamble now emits each user import exactly once. Previously the `<script setup>` imports went through the deduped block while the preserved normal `<script>` body was appended verbatim with its own `import` statements still inside, so a shared specifier produced duplicate import statements in the compiled module - duplicate named bindings ("already declared") in the strict case, and double-running top-level side effects in the case #993 describes.

Two coordinated changes in `compile_script/inline/compiler/preamble.rs`:
- `normal_script_imports` are merged into the deduped emit (`dedupe_imports` over setup + normal imports) instead of only feeding the template-binding map, so a normal-script-only import (including bare side-effect imports) still lands exactly once
- the appended normal-script body goes through `strip_import_statements`, which mirrors `parser::parse_script_content`'s import detection (single-line side-effect imports, multi-line imports, template-literal guard) and keeps every non-import statement verbatim

## Context

#993 traces the duplication to the exact two sites changed here: the deduped import emit (which only covered `<script setup>` imports) and the verbatim append of the preserved normal `<script>` body. Any top-level side effect in a doubly-imported module - registering a hotkey listener, pushing into a global registry, calling `installPlugin()` - fired twice in the compiled module, and shared named imports produced duplicate bindings.

## Change Class

- [x] Compiler and codegen

## Behavior Reference

Fixes #993 (code-grounded issue; the L118-128/L150 sites it cites are exactly the two changed). Matches `@vue/compiler-sfc`, which hoists and dedupes user imports from both blocks ahead of the component definition.

## Verification Evidence

Commands: `cargo test -p vize_atelier_sfc` (454 passed), `cargo check -p vize_atelier_sfc`, `cargo fmt --check`.

- [x] Minimal fixture or focused regression test added or updated - new `test_compile_both_script_blocks_dedupes_imports` asserts the issue's exact fixture shape: shared `import { ref } from "vue"` in both blocks emitted once, normal-only `registerHotkey` import emitted once, bare `import "./side-effect-only"` survives exactly once, and the `registerHotkey()` call still flows through
- [x] Snapshot/baseline changes reviewed and explained - three snapshots refresh with whitespace-only diffs (blank-line placement around the emitted import block); no content changes, reviewed line by line
- [x] Parity or real-world fixture coverage considered - dual-block fixtures `compile_both_script_blocks*` and `script_setup__script_*` all pass; `<script setup>`-only output is byte-identical (strip path only runs when a normal block exists)
- [ ] Security/audit impact considered - n/a, codegen ordering only
- [ ] Performance status or PR benchmark impact considered - one extra line scan over the (small) preserved normal script per SFC
- [ ] Fuzzing target or crash-reproducer coverage considered - n/a
- [ ] Editor/LSP scenario coverage considered - n/a, inline compile output only
- [ ] Ecosystem or broad app compatibility impact considered - output for single-block SFCs is unchanged
- [ ] Docs, release, or stability notes updated when public behavior changed - no public API change

## Risk

Low. The strip helper intentionally mirrors the existing parser's import detection so the two stay in agreement; if they ever diverge, the worst case is an import emitted twice (the status quo this PR fixes), never a dropped statement, because stripping only removes lines the parser already classified as imports.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/997"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783172961&installation_id=136613492&pr_number=997&repository=ubugeeei-prod%2Fvize&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fvize%2Fpull%2F997&signature=5d62c2ef5fdcd86da644c8a6b7b415eb48d533cd97bbe303973dd0f9bafbf5b7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/crates/vize_atelier_sfc/src/compile/snapshots/vize_atelier_sfc__compile__tests__compile_both_script_blocks.snap
+++ b/crates/vize_atelier_sfc/src/compile/snapshots/vize_atelier_sfc__compile__tests__compile_both_script_blocks.snap
@@ -1,13 +1,12 @@
 ---
 source: crates/vize_atelier_sfc/src/compile/tests.rs
-assertion_line: 1165
 expression: result.code.as_str()
 ---
 import { defineComponent as _defineComponent } from 'vue'
 import { toDisplayString as _toDisplayString, openBlock as _openBlock, createElementBlock as _createElementBlock, Fragment as _Fragment, renderList as _renderList } from "vue"
 
-
 import type { RouteLocation } from "vue-router";
+
 
 interface TabItem {
   name: string;
--- a/crates/vize_atelier_sfc/src/compile/snapshots/vize_atelier_sfc__compile__tests__compile_both_script_blocks_exposes_normal_script_bindings_to_template.snap
+++ b/crates/vize_atelier_sfc/src/compile/snapshots/vize_atelier_sfc__compile__tests__compile_both_script_blocks_exposes_normal_script_bindings_to_template.snap
@@ -5,8 +5,8 @@ expression: result.code.as_str()
 import { defineComponent as _defineComponent } from 'vue'
 import { toDisplayString as _toDisplayString, normalizeClass as _normalizeClass, openBlock as _openBlock, createElementBlock as _createElementBlock, Fragment as _Fragment, renderList as _renderList } from "vue"
 
-
 import { ref } from "vue";
+
 
 const dragging = ref(false);
 let dropCallback = null;
--- a/crates/vize_atelier_sfc/src/compile/tests.rs
+++ b/crates/vize_atelier_sfc/src/compile/tests.rs
@@ -1166,6 +1166,65 @@ const { items } = defineProps<{
 }
 
 #[test]
+fn test_compile_both_script_blocks_dedupes_imports() {
+    // Regression test for #993: an import present in BOTH blocks must be
+    // emitted exactly once, and a side-effect import that lives only in the
+    // normal `<script>` must survive (exactly once).
+    let source = r#"<script lang="ts">
+import { ref } from "vue";
+import { registerHotkey } from "./hotkey-plugin";
+import "./side-effect-only";
+
+registerHotkey();
+</script>
+
+<script setup lang="ts">
+import { ref } from "vue";
+
+const count = ref(0);
+</script>
+
+<template>{{ count }}</template>"#;
+
+    let descriptor = parse_sfc(source, SfcParseOptions::default()).expect("Failed to parse SFC");
+    let opts = SfcCompileOptions {
+        script: ScriptCompileOptions {
+            is_ts: true,
+            ..Default::default()
+        },
+        template: TemplateCompileOptions {
+            is_ts: true,
+            ..Default::default()
+        },
+        ..Default::default()
+    };
+    let result = compile_sfc(&descriptor, opts).expect("Failed to compile SFC");
+    let code = result.code.as_str();
+
+    assert_eq!(
+        code.matches("import { ref } from \"vue\"").count(),
+        1,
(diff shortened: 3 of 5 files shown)


ubugeeei:
Ah, sorry about this — turns out we were both tackling the exact same issue at the same time, and a parallel fix for #993 (dedupe imports across both blocks + strip the normal `<script>` body's own imports) already went in. Really bad timing on our side! 🙏

Thank you so much for the contribution and the thorough write-up — it's genuinely appreciated, and we'd love to see more PRs from you! 🎉

PR #1065, opened 2026-06-06
fix(parser): report stricter HTML parse errors

ubugeeei:
## Summary
- report/recover more HTML tokenizer parse errors: processing instructions, unexpected solidus, missing attribute values, duplicate attributes ignoring ASCII case, nested comments
- align tree construction with browser-like HTML behavior for non-void self-closing tags, implicit table sections, foster parenting, in-body optional closures, nested form/a/button handling, and a focused adoption-agency repair path
- update downstream fixtures/snapshots that relied on self-closing non-void HTML elements, and adjust permitted-content lint tests for parser-repaired trees

## Validation
- cargo test --workspace

## Notes
- Vue directives, components, and mustache syntax remain supported. Component self-closing syntax is kept where the parser classifies the tag as a component.
- SFC templates are parsed as fragments, so document-level quirks mode/doctype state is not represented in the AST in this PR.

### DIFF
--- a/crates/vize/tests/build_cli.rs
+++ b/crates/vize/tests/build_cli.rs
@@ -60,7 +60,7 @@ withDefaults(defineProps<AppProps>(), {
     :data-as="as"
     :data-as-child="asChild"
     :data-feature="feature"
-  />
+  ></div>
 </template>
 "#,
     );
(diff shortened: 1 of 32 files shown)


sevenc-nanashi:
Could it be reverted? While this does not comply HTML spec, but Vue encourages using `<div />`, at least in [their linter](https://eslint.vuejs.org/rules/html-self-closing).

ubugeeei:
This is an parser-related issues.

follow up: https://github.com/ubugeeei-prod/vize/pull/1085

PR #1095, opened 2026-06-06
perf(atelier): slice the render body via codegen offsets instead of re-scanning

ubugeeei:
`extract_template_parts` re-discovered render-body boundaries that codegen already knew: it ran **two byte-level state machines** (brace + paren tracking, `count_braces_with_state` / `count_parens_with_state`) over the generated render function and **re-copied the body line by line**, purely to split out a slice the compiler had just produced.

Now codegen records the render body's byte span (right after `return ` up to the closing `}`) on `CodegenResult.render_body_span`. The SFC layer adjusts it for the `preamble + "\n"` prefix and threads it to `extract_template_parts`, which:
- slices the render body directly (`&code[start..end]` + the existing trailing trim), and
- classifies only the surrounding lines (imports / hoisted / component-directive preamble / render fn name) with cheap `starts_with` checks — no state machine.

The line scanner is **kept as a fallback** for the `function render($props)` block-render shape, out-of-bounds/non-char-boundary spans, and any output without a recognized render fn (SSR/vapor pass `None` and use `extract_template_parts_full`).

- **Output byte-identical**: 704 atelier tests green, including the `extract_template_parts*` snapshots that pin the slice-vs-scanner equivalence.
- **`atelier.sfc.template.extract_parts` drops 9.5ms → 1.7ms (-82%)**; full `atelier.sfc.compile` ~101ms → ~89ms on a binding-heavy component (`vize build --profile`).

From the profile-guided whole-pipeline audit (rank #1). fmt/clippy(-D warnings) clean.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/1095"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783355572&installation_id=136613492&pr_number=1095&repository=ubugeeei-prod%2Fvize&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fvize%2Fpull%2F1095&signature=028119ba5a2ad82a8109864f28a41dffeb6fb1c42c99b9428c8afd649b8a8374"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/crates/vize_atelier_core/src/codegen.rs
+++ b/crates/vize_atelier_core/src/codegen.rs
@@ -59,6 +59,11 @@ pub fn generate(root: &RootNode<'_>, options: CodegenOptions) -> CodegenResult {
     // Generate return statement
     ctx.push("return ");
 
+    // Record where the render body begins so the SFC layer can slice it out
+    // instead of re-scanning the generated function. `ctx.code` is appended to
+    // monotonically from here, so this byte offset stays valid.
+    let render_body_start = ctx.code.len();
+
     // Generate root node
     if root_children.is_empty() {
         ctx.push("null");
@@ -105,6 +110,10 @@ pub fn generate(root: &RootNode<'_>, options: CodegenOptions) -> CodegenResult {
             ctx.push("], 64 /* STABLE_FRAGMENT */))");
         }
     }
+
+    // Record the end of the render body (before the closing brace's leading
+    // newline + `}`), completing the slice the SFC layer can lift directly.
+    let render_body_end = ctx.code.len();
 
     ctx.deindent();
     ctx.newline();
@@ -156,6 +165,7 @@ pub fn generate(root: &RootNode<'_>, options: CodegenOptions) -> CodegenResult {
         code: ctx.into_code(),
         preamble,
         map: None,
+        render_body_span: Some((render_body_start, render_body_end)),
     }
 }
 
--- a/crates/vize_atelier_core/src/codegen/context.rs
+++ b/crates/vize_atelier_core/src/codegen/context.rs
@@ -74,6 +74,11 @@ pub struct CodegenResult {
     pub preamble: String,
     /// Source map (JSON)
     pub map: Option<String>,
+    /// Byte offsets `(start, end)` of the render body within `code` — the slice
+    /// immediately after `return ` up to (but excluding) the closing `}`. Lets
+    /// the SFC layer slice the render body out instead of re-scanning and
+    /// re-copying it line by line. `None` when the backend does not record it.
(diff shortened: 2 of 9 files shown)


ubugeeei:
⚠️ **Blocked by `cargo-semver-checks (vize_atelier_core)`** — marking draft for a maintainer API decision.

The change adds a `pub` field (`render_body_span`) to the `pub struct CodegenResult`, which is a breaking change (it stops external code from constructing the struct with a literal). The optimization itself is verified and works:

- `atelier.sfc.template.extract_parts`: **9.5ms → 1.7ms (-82%)**
- full `atelier.sfc.compile`: ~101ms → ~89ms on a binding-heavy component
- output byte-identical (704 atelier tests incl. `extract_template_parts*` snapshots green); fmt/clippy clean

To land semver-clean, the render-body span needs an API-compatible carrier. Options (maintainer's call):
1. `#[non_exhaustive]` on `CodegenResult` (one-time major) so future fields are non-breaking, then add the field;
2. add a new `generate_with_render_span()` in atelier_core + a matching atelier_dom entry (new pub fns are non-breaking), leaving `CodegenResult`/existing fns untouched;
3. accept a major bump for atelier_core.

Happy to implement whichever you prefer.

ubugeeei:
Closing. Local re-measurement on 300 typical files shows this is **neutral**, not a regression (baseline 34.0ms vs this branch 34.1ms min over 9 runs) — the CI Compile-SFC +5.90% was small-corpus benchmark **noise** (same pattern as earlier noise on a maestro-only PR).

But it's not a clean win worth landing: (1) semver-blocked (adds a `pub` field to `pub CodegenResult`), and (2) the real gain (`extract_parts` -82%) only matters on very large render bodies — on typical components `extract_parts` is a tiny fraction of compile, so the net is noise. Parking the idea; if wanted later it should land via a semver-clean carrier (e.g. a new `generate_with_render_span` API), and ideally only when the broader compile path makes it measurable.

PR #1119, opened 2026-06-07
feat(legacy): type literal Vue 2 data() bindings in the virtual TS

ubugeeei:
## What

Deepens opt-in legacy **Vue 2.7 / Nuxt 2** Options API type-checking: a `data()` property initialized with a literal now gets a precise TypeScript type in the generated virtual TS instead of the blanket `any`.

```vue
<script>
export default { data() { return { count: 1, label: 'x', flag: true } } }
</script>
```

Virtual TS, before → after:

```diff
- const count: any = undefined as any;
- const label: any = undefined as any;
- const flag: any = undefined as any;
+ const count: number = undefined as any;
+ const label: string = undefined as any;
+ const flag: boolean = undefined as any;
```

So the type checker can now catch e.g. `{{ count.toUpperCase() }}` in a Vue 2 template. Inference is sound-by-construction: only number / string / boolean / bigint / substitution-free template literals are typed; **anything else falls back to `any`**, so this is a strict superset of the previous behavior and can never introduce a false template type error. (`computed` / `methods` return-type inference are natural follow-ups built on the same plumbing.)

## Zero-cost for the default Vue 3 build

The whole feature is gated behind the `legacy` cargo feature and **compiled out** of the standard pipeline — no runtime branch, no struct growth, no memory on the Vue 3 path:

- New `vize_croquis` `legacy` feature (wired from `vize_canon/legacy`).
- The type side-table (`BindingMetadata::legacy_template_binding_types`), the inference (`infer_legacy_literal_ts_type`), the capture in `data()` collection, and the precise-type emission in the virtual-TS generator are **all `#[cfg(feature = "legacy")]`** — they do not exist in the default binary.
- `BindingMetadata`'s `Debug` is hand-written to omit the legacy-only side-table, so the field is invisible to snapshots/diagnostics and the default and legacy builds produce identical debug output.

## Verification

- **Default (Vue 3): zero regression** — `cargo test -p vize_croquis -p vize_canon -p vize_atelier_sfc` all green, output byte-identical (the new code is cfg'd out).
- **Legacy** — `cargo test -p vize_croquis -p vize_canon --features legacy` green, incl. two new tests: literal inference in `vize_croquis`, and end-to-end `const count: number` in the generated virtual TS (`vize_canon`).
- `cargo clippy --features legacy -- -D warnings` clean; `cargo fmt --check` clean; `vize` CLI builds with and without `--features legacy`.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/1119"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783432877&installation_id=136613492&pr_number=1119&repository=ubugeeei-prod%2Fvize&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fvize%2Fpull%2F1119&signature=e64aa2e2f10e163aff07c8c730b56271db2ac385c6203d5e56dd2bc84a3db467"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/crates/vize_canon/Cargo.toml
+++ b/crates/vize_canon/Cargo.toml
@@ -12,7 +12,7 @@ default = ["native"]
 native = ["dep:which", "dep:walkdir", "dep:dirs", "dep:corsa", "dep:lsp-types"]
 # Opt-in legacy Vue (v0 / v1 / v2) type-checking surface. Not enabled by the
 # default Vue 3 build; turns on the parser-side legacy feature too.
-legacy = ["vize_armature/legacy"]
+legacy = ["vize_armature/legacy", "vize_croquis/legacy"]
 
 [dependencies]
 vize_carton.workspace = true
--- a/crates/vize_canon/src/sfc_typecheck.rs
+++ b/crates/vize_canon/src/sfc_typecheck.rs
@@ -463,6 +463,38 @@ export default {
         );
     }
 
+    #[cfg(feature = "legacy")]
+    #[test]
+    fn test_legacy_vue2_data_literal_types_in_virtual_ts() {
+        let source = r#"<script lang="ts">
+export default {
+  data() {
+    return { count: 1, label: 'x', flag: true, other: window.foo }
+  }
+}
+</script>
+<template>
+  <div>{{ count }} {{ label }} {{ flag }} {{ other }}</div>
+</template>"#;
+
+        let mut options = SfcTypeCheckOptions::new("Legacy.vue");
+        options.include_virtual_ts = true;
+        let result = type_check_sfc_with_legacy_vue2(source, &options);
+        let ts = result
+            .virtual_ts
+            .expect("virtual TypeScript should be generated");
+
+        // Literal `data()` initializers now carry precise types instead of `any`,
+        // so the type checker can flag misuse (e.g. `count.toUpperCase()`).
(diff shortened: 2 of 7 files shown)


ubugeeei:
Closing: under the confirmed 3-tier taxonomy, Options API is NOT legacy (it's officially v3-supported, opt-in + documented). Gating data() typing behind the `legacy` cargo feature is the wrong home, and adding a pub field to BindingMetadata trips cargo-semver-checks. This will be re-landed as part of decoupling Options API into a standard-binary config opt-in (semver-safe, no new pub struct field).

Issue #1352, opened 2026-06-10
Pass through the gates of Production Ready with Fable.


ubugeeei:
A single tracking issue for making the following 9 components production ready, validated against real-world projects (Misskey / VOICEVOX / Elk / Vue Vben Admin / Hoppscotch / Element Plus / Ant Design Vue / Reka UI / nuxt-ui / npmx / vuefes / frontend-phpcon).

### 1. Compiler (vize_armature / vize_croquis / vize_atelier_*)
- [x] `cargo test --workspace` green (3674 passed / 0 failed)
- [x] Real-world build snapshots green (8/8: elk / misskey / npmx / vuefes production builds)
- [x] Inspector parity budgets green (elk / npmx / vuefes; refreshed in #1358 after #1349)
- [x] **Fixed:** generic-interface inherited props dropped (#1357, merged) — root cause of nuxt-ui SSR 500
- [x] **Fixed:** types not collected through plain `export *` barrels (#1362) — Button lost `Omit<LinkProps,...>` props
- [ ] Compiler fixture budget 625/625 (verified via coverage run — pending final cross-cutting pass)

### 2. Vite Plugin (@vizejs/vite-plugin)
- [x] Preview smoke tests green (4/4 apps)
- [x] nuxt-ui dev SSR verified locally: home page & button page render (was 500); needs nightly e2e to confirm post-merge
- [ ] Full e2e dev suite + VRT green on nightly after open PRs merge

### 3. Nuxt Integration (@vizejs/nuxt)
- [x] nuxt-ui dev (Nuxt 4 + @nuxt/content + vize module) renders SSR pages locally with open fixes
- [ ] Nuxt e2e (nuxt-ui / npmx / vuefes / frontend-phpcon) green on nightly

### 4. Type Checker (vize_canon)
- [x] Fixture check snapshots 10/10 (incl. vue-tsc parity)
- [x] Real-world check snapshots 11/11 (stale tsgo pin fixed; baselines refreshed in #1364 after #1357)
- [x] `vize check` crash-free on all 11 real-world projects

### 5. Linter (vize_patina / oxlint-plugin-vize)
- [x] **Fixed:** multibyte panic in `no_hydration_mismatch` (#1353) — found on Misskey
- [x] Lint snapshots 11/11, crash-free on all fixtures
- [x] oxlint bridge example verified (clean pass + intentional failure-case)

### 6. Formatter (vize_glyph)
[...]

PR #1371, opened 2026-06-10
fix(canon): resolve editor imports against the real workspace

ubugeeei:
## Summary
User report: the editor shows **`Cannot find module`** type errors on `<script>` imports in `.vue` files — including plain relative imports like `../types` — while `vize check` reports nothing.

Root cause (two stacked defects in the LSP→corsa path):
1. The corsa session behind the LSP bridge was rooted in an **isolated scratch directory** (`CorsaProjectClient::new`), not the workspace.
2. `build_session_document_uri` kept a document's real path only when the path **exists on disk** — a virtual `.vue.ts` overlay never does — so every open document was remapped into an overlay mirror where none of the project's real `.ts` neighbors exist. TypeScript then failed every relative import (and the project's tsconfig `paths` never applied).

Fix:
- Root the bridge session at the workspace whenever one is known (project tsconfig drives resolution, matching `vize check`); the scratch session remains the rootless fallback.
- Keep `.vue.ts`/`.html.ts` overlays whose **underlying real file** exists at their real path. These sync as pure in-memory overlays (`refresh_with_overlay_changes`) — nothing is written into the user's source tree.

## Verification
- LSP harness against nuxt-ui `Button.vue` (typecheck enabled): `Cannot find module` diagnostics **5 → 0** (parity with the CLI baseline, which reports none for relative imports)
- New unit test: in-project `.vue.ts` overlay keeps its real path; out-of-project paths still remap
- `cargo test -p vize_canon -p vize_maestro`: green; clippy clean

Part of #1352 (Production Gate).

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/1371"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783688812&installation_id=136613492&pr_number=1371&repository=ubugeeei-prod%2Fvize&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fvize%2Fpull%2F1371&signature=9a217375b0858fb6005caaecbdc418c06f772efd783c90ff886dfce3888314f7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/crates/vize_canon/src/corsa_bridge/bridge.rs
+++ b/crates/vize_canon/src/corsa_bridge/bridge.rs
@@ -74,8 +74,20 @@ impl CorsaBridge {
             .as_ref()
             .map(|path| path.to_string_lossy().into_owned());
 
-        let client = CorsaProjectClient::new(corsa_path.as_deref(), working_dir.as_deref())
-            .map_err(CorsaBridgeError::SpawnFailed)?;
+        // Root the session at the real workspace whenever one is known: the
+        // project's own tsconfig (paths, baseUrl) then drives module
+        // resolution and virtual `.vue.ts` overlays can live at their real
+        // paths, so relative imports in `<script>` resolve exactly like
+        // `vize check`. The isolated scratch session remains the fallback
+        // for rootless (single-file) usage.
+        let client = match working_dir.as_deref() {
+            Some(dir) => CorsaProjectClient::new_for_workspace(
+                corsa_path.as_deref(),
+                std::path::Path::new(dir),
+            ),
+            None => CorsaProjectClient::new(corsa_path.as_deref(), None),
+        }
+        .map_err(CorsaBridgeError::SpawnFailed)?;
         *guard = Some(client);
 
         self.initialized.store(true, Ordering::SeqCst);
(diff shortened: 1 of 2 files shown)


ubugeeei:
⚠️ Found a regression after merge while extending verification: with the workspace-rooted session, a virtual `.vue.ts` overlay is **not part of the tsgo program** (it doesn't exist on disk and tsconfig only includes real files), so corsa returns zero diagnostics — `lsp_corsa_smoke_publishes_diagnostics_and_hover` fails locally (it appears to be skipped on CI when the corsa runtime is unavailable). The false `Cannot find module` storm is gone, but so is editor type checking. Working on the proper fix now (making the overlay a program member); will have a follow-up PR shortly.

Issue #1629, opened 2026-06-13
feat(patina): port uncovered rules from eslint-plugin-{react,vue,svelte,vuejs-a11y}


ubugeeei:
## Overview

Tracking issue for expanding `vize_patina` lint rules by porting still-relevant rules from the major ESLint plugin ecosystems. The goal is to **implement every rule that is still valid for modern Vue 3 / modern JSX**, surveyed across:

- [jsx-eslint/eslint-plugin-react](https://github.com/jsx-eslint/eslint-plugin-react)
- [vuejs/eslint-plugin-vue](https://github.com/vuejs/eslint-plugin-vue)
- [sveltejs/eslint-plugin-svelte](https://github.com/sveltejs/eslint-plugin-svelte)
- [vue-a11y/eslint-plugin-vuejs-accessibility](https://github.com/vue-a11y/eslint-plugin-vuejs-accessibility)

> [!NOTE]
> Filed as a single tracking issue. Milestones (M1–M7) are intended to be knocked out via incremental PRs.

> [!IMPORTANT]
[...]

PR #1700, opened 2026-06-13
feat(patina): add vue/no-v-for-template-key-on-child

ubugeeei:
Implements `vue/no-v-for-template-key-on-child` from #1629.

In Vue 3 the `key` for a `<template v-for>` must live on the `<template>` element itself, not on its child (the reverse of Vue 2). This rule reports when a `<template v-for="...">` has a direct child element carrying a `:key`/`v-bind:key`, so the key can be moved up onto the template.

Mirrors eslint `vue/no-v-for-template-key-on-child` (the Vue 3 variant). **essential** preset.

## Implementation
- Rule impl: `crates/vize_patina/src/rules/vue/no_v_for_template_key_on_child.rs` (LintContext `enter_element` walks the `<template v-for>` children and reports any bound `:key` on a direct child element).
- Registered via the shared growable `register_valid_directives` bundle in `rules/vue.rs` (called from both `with_essential` and `with_happy_path`), so `rule.rs` is untouched per the source-length guard.
- i18n (en/ja/zh) added to `i18n_supplemental_extra.rs`.
- Updated the preset-membership snapshot and `npm/vize/src/types/rules.ts`.

## Verification
`cargo test -p vize_patina` is green (1657 passed, 0 failed), including 7 new inline tests, registry consistency, and the regenerated membership snapshot.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/1700"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a>


ubugeeei:

PR #1701, opened 2026-06-13
feat(patina): add script/valid-define-props

ubugeeei:
Implements `script/valid-define-props` from #1629.

Validates `defineProps` macro usage in `<script setup>` (mirrors eslint [vue/valid-define-props](https://eslint.vuejs.org/rules/valid-define-props.html)), modeled on the existing `script/valid-define-options` rule:

- ERROR when `defineProps` has **both** a type argument and a runtime argument (`defineProps<{...}>({...})`).
- ERROR when the runtime argument **references locally-declared variables** (a hoisted `defineProps` cannot reference module-local bindings).
- ERROR on **multiple** `defineProps` calls per block.

Registered in the growable script-rule registry (`registry/names.rs`, `registry/rules.rs`) with `category: Script` and the opinionated/nuxt presets, matching the sibling `valid-define-options`. Membership snapshot and `rules.ts` updated; inline tests cover valid + invalid cases (15 tests). `cargo test -p vize_patina` is green (1665 passed).

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/1701"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a>


ubugeeei:

PR #1702, opened 2026-06-13
feat(patina): add script/valid-define-emits

ubugeeei:
Implements `script/valid-define-emits` from #1629.

Validates `defineEmits` macro usage in `<script setup>` (mirrors eslint `vue/valid-define-emits`), reporting an error when the macro:

- is given **both** a type argument and a runtime argument (mutually exclusive),
- **references locally-declared variables** in its runtime argument (the macro is hoisted above setup), or
- is **called more than once** per block.

Modeled exactly after `script/valid-define-options` (#1691): same `ScriptRule` AST-visitor structure, diagnostic emission, registry wiring (`names.rs` / `rules.rs`, growable script-rule registry), and `Script` category / opinionated preset registration. Per #1629 (M4) the rule is tracked under the `essential` category; following the reference rule it ships in the same `Script`/opinionated slot.

Inline tests cover valid runtime/type-only forms, imported-binding references, and each invalid case (type+runtime, local var / function / destructured-local reference, duplicate call), plus offset handling. `cargo test -p vize_patina` is green (1662 tests).

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/1702"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a>


ubugeeei:

PR #1709, opened 2026-06-13
feat(patina): add vue/valid-template-root

ubugeeei:
Implements `vue/valid-template-root` from #1629.

Validates the SFC `<template>` root for Vue 3 fragment semantics. Reports a root-level `<template>` or `<slot>` element used as a root node — neither is a renderable element, so neither can stand in as a root. Vue 3 allows multiple root nodes and `v-if`/`v-else` roots, so those are not flagged.

### Scope (sound subset)
eslint's rule also reports an empty `<template>` and `v-for` on a root node. In patina a template root is linted identically whether it is a full SFC root or an isolated fragment/snippet (an empty fragment and a single `v-for` root are both valid inputs — see the existing `test_lint_empty_template` / `test_lint_valid_template` contracts), so those two checks cannot be expressed soundly here and would produce false positives. Per the issue's guidance, this ports the sound subset: the root-element-kind check, which holds regardless of context. `essential` preset (modeled after the #1633 `valid-v-*` rules; registered via the shared `register_valid_directives` bundle).

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/1709"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a>


ubugeeei:

PR #1711, opened 2026-06-13
feat(patina): add script/require-explicit-slots

ubugeeei:
Implements `script/require-explicit-slots` from #1629.

## What it does

In a TypeScript `<script setup>`, slots that are consumed in script should be explicitly typed via `defineSlots<...>()`. This rule flags the clear case where slots are consumed via `useSlots()` but there is **no** `defineSlots(...)` / `defineSlots<...>()` declaration in the same block. Mirrors [`vue/require-explicit-slots`](https://eslint.vuejs.org/rules/require-explicit-slots.html) (opinionated preset).

Reported (warning), anchored at the first `useSlots()` call:
```ts
const props = defineProps<{ id: number }>()
const slots = useSlots()   // ← Slots consumed via useSlots() must be explicitly typed with defineSlots<...>().
```

Not reported:
```ts
defineSlots<{ default(props: { msg: string }): unknown }>()
const slots = useSlots()
```

## Scope (conservative subset)

This is a `script/*` rule operating on a single script block, so it implements the **soundest subset** of the eslint semantics that is observable from script alone, prioritising zero false positives:

- **TS-only gate.** The eslint rule early-exits unless a `<script>` has `lang="ts"`. A patina script rule does not receive the SFC `lang` attribute, so the presence of any TypeScript syntax in the block (a type annotation, interface/type alias, `defineProps<T>()`, a `defineSlots<T>()` type argument, etc.) is used as a sound proxy for `lang="ts"`. A block with **no** TS syntax is treated as JavaScript and never flagged — in JS the type-only `defineSlots<T>()` fix is unavailable, so flagging would be a false positive. (Trade-off: a TS block whose *only* content is `useSlots()` with no other TS token is not flagged — an accepted false negative in exchange for soundness.)
- **Script-side consumption only.** Upstream `vue/require-explicit-slots` also validates `<slot>` elements in the template against the declared slots. That is template/cross-block analysis outside a single script block; this port covers the `useSlots`-without-`defineSlots` portion, which is what a script rule can observe soundly.
- Any `defineSlots(...)` call (typed or not) in the block satisfies the rule.

The behaviour is modeled on the sibling TS/slots script rules `script/require-typed-ref` (TS-only gating, AST visitor) and `script/prefer-use-slots` (useSlots/$slots detection): same `ScriptRule` shape, registered alongside the other opinionated script rules.

## Tests

13 inline `#[cfg(test)]` cases (valid + invalid): useSlots without defineSlots across several TS signals (type annotation, interface, `defineProps<T>`), `defineSlots<T>()` satisfying, `defineSlots` declared before/after useSlots, JS blocks (no TS syntax) not flagged, runtime `defineProps([...])` staying JS, string-literal false-positive guard, multiple-useSlots reports once at the first, and offset handling.

## Verification

- `cargo test -p vize_patina` — green (1663 passed).
- Preset membership insta snapshot and `npm/vize/src/types/rules.ts` regenerated.
- Inline messages only; no i18n table change.
- `rustfmt` (toolchain 1.95.0 / rustfmt 1.9.0) clean on every touched `.rs`; size guard respected (no edited source file > 350 lines; `i18n_supplemental.rs` unchanged at 347; `rule.rs` / `script_rules.rs` untouched).


ubugeeei:

PR #1712, opened 2026-06-13
feat(patina): add vue/no-multiple-slot-args

ubugeeei:
Implements `vue/no-multiple-slot-args` from #1629.

## What

Adds the script-side rule `script/no-multiple-slot-args` (preset: `opinionated`/`nuxt`, category `Script`). It disallows passing more than one argument — or a spread argument — to a scoped-slot function call, because Vue forwards only a single "slot props" value to the slot binding; any further positional arguments are silently dropped.

Ports [`vue/no-multiple-slot-args`](https://eslint.vuejs.org/rules/no-multiple-slot-args.html), extended to the Composition-API forms.

### Reported (warning)

```ts
slots.default(foo, bar)        // Unexpected multiple arguments
$slots.header(a, b)
$scopedSlots.item(a, b)
this.$scopedSlots.foo(a, b)    // Options API render function
this.$slots.foo(a, b)
useSlots().default(a, b)
slots['default'](a, b)         // string-literal computed slot name
slots.default?.(a, b)          // optional call
slots.default(...args)         // Unexpected spread argument
```

### Allowed

```ts
slots.default({ foo, bar })    // single slot-props object
slots.default(slotProps)
slots.default()
```

## Scope (sound conservative subset)

eslint's rule only covers the Options-API `this.$slots` / `this.$scopedSlots` render-function forms. Since patina targets modern `<script setup>`, this implementation additionally recognises the Composition-API slot sources, matching the rule note in #1629 (`slots.default(a, b)`, `$slots.foo(a, b)`, `useSlots().foo(a, b)`).

A call is flagged only when its callee is a member access **rooted in a recognised slots source**:

- a bare `slots` / `$slots` / `$scopedSlots` identifier,
- `this.$slots` / `this.$scopedSlots`,
- the result of a `useSlots()` call.

To minimise false positives, matching is intentionally narrow:

- The slot name must be a static identifier or a **string-literal** computed key; a dynamic computed key (`slots[name](a, b)`) is treated as not-statically-a-slot and left alone.
- An arbitrary multi-argument method call on an unrelated object (`foo.bar(a, b)`, `emit('x', y)`, `Math.max(a, b)`) is never reported.
- A bare `slots(a, b)` (not a member call) is not a slot invocation and is ignored.

This means data-flow aliasing (`const f = this.$scopedSlots.foo; f(a, b)`) is out of scope — tracking it soundly would require reference analysis and risk false positives. A correct narrow rule was preferred over an unsound broad one.

## Tests

26 inline `#[cfg(test)]` cases (valid + invalid), covering every recognised slots source, the spread form, computed string keys, optional calls, offset handling, multiple occurrences, and the conservative non-matches. Two insta snapshots assert the multiple-args and spread diagnostics.

## Verification

- `cargo test -p vize_patina` — green (1676 tests).
- `rustfmt --edition 2024 --check` (toolchain 1.95.0) — clean on every touched file; no unrelated formatting drift.
- `cargo clippy -p vize_patina` — clean.
- Source-length guard — every touched file is well under 350 lines; `i18n_supplemental.rs` untouched (no i18n needed — inline messages, consistent with the other `script/*` validation rules).
- Regenerated the preset membership snapshot and `npm/vize/src/types/rules.ts`.


ubugeeei:

PR #1713, opened 2026-06-13
feat(patina): add script/no-duplicate-attr-inheritance

ubugeeei:
Implements `script/no-duplicate-attr-inheritance` from #1629.

Ports the redundant-by-default core of eslint's [`vue/no-duplicate-attr-inheritance`](https://eslint.vuejs.org/rules/no-duplicate-attr-inheritance.html) (opinionated preset; registered into `opinionated` + `nuxt`).

## Scope (sound conservative subset)

eslint's rule reports *double attribute application*: a component that keeps the default attribute inheritance (`inheritAttrs: true`) **and** also forwards `$attrs` manually (typically `v-bind="$attrs"` in the template). Detecting that requires pairing a **template** signal (`v-bind="$attrs"`) with a **script** signal (the effective `inheritAttrs` value).

In the patina architecture markup-rules and script-rules are separate passes — a script rule cannot observe the template's `v-bind="$attrs"`. Rather than approximate that cross-block relationship unsoundly (which would false-positive on the many components that set `inheritAttrs: true` but do *not* forward `$attrs`), this rule implements the **template-independent subset** that is always correct:

> An **explicit** `inheritAttrs: true` is redundant, because `true` is the framework default. Removing it changes nothing.

This is exactly the location eslint also points at when both signals are present, and because `true` is unconditionally the default the rule fires with **zero false positives** — every flagged occurrence is a genuine no-op option. Components that opt out with `inheritAttrs: false`, or that omit the option entirely, are never touched. The template-side pairing (`v-bind="$attrs"` with an *implicit* default) is intentionally out of scope and can be added later as a markup rule once cross-block analysis is wired up.

## What it flags

An explicit `inheritAttrs: true` resolved via the same machinery as `script/component-options-name-casing`:

- `<script setup>`: `defineOptions({ inheritAttrs: true })`
- Options API: `export default { inheritAttrs: true }`, `defineComponent({ inheritAttrs: true })`, or an identifier bound to such an object (unwrapping TS `as` / `satisfies` / non-null / parenthesized wrappers).

Only a literal `true` is reported; `inheritAttrs: false`, a computed key, or any non-boolean-literal value (an identifier, a call, …) is left alone so the rule never guesses at a value it cannot see.

## Tests

15 inline `#[cfg(test)]` cases (valid + invalid), covering `defineOptions`, `export default`, `defineComponent`, identifier-bound options, string-literal keys, the `satisfies` wrapper, `inheritAttrs: false`, missing option, non-literal value, computed key, and offset handling. One insta snapshot for the primary diagnostic. Preset membership snapshot and `npm/vize/src/types/rules.ts` regenerated.


ubugeeei:

PR #1714, opened 2026-06-13
feat(patina): add vue/no-deprecated-functional-template

ubugeeei:
Implements `vue/no-deprecated-functional-template` from #1629 (M5: Vue 2 → 3 migration pack, opt-in).

## What it does

Reports the Vue 2 `functional` attribute on the SFC `<template>` block. Functional single-file-component templates (`<template functional>`) were removed in Vue 3 — the attribute is now silently ignored, so the migration intent breaks. Functional components in Vue 3 are written as plain functions (render function / JSX), not as SFC templates. This mirrors eslint-plugin-vue's rule of the same name. It is **opt-in** (registered only via `rules::vue::register_opt_in`, not in any preset) and gates itself on the default Vue 3 dialect, following the sibling `no-deprecated-*` migration rules.

## Scope / architecture note

patina runs markup rules and SFC-level rules in **separate passes**. The SFC `<template>` opening tag (and its attributes) is not part of the template AST handed to markup rules — `extract_template_fast` / the descriptor template-parse only feed the block's *inner content* to the markup pass, so a markup-AST rule (`enter_element`) can never observe the root `<template functional>`.

The `functional` attribute is, however, preserved on the parsed SFC descriptor's template block (`descriptor.template.attrs`). The rule therefore runs as an SFC-level rule (`run_on_sfc`) over that descriptor — reusing the engine's shared descriptor when present and otherwise parsing the SFC itself, exactly like `vue/no-empty-component-block` and `vue/single-style-block`. This is precisely the SFC root `<template>` element eslint-plugin-vue targets (a nested `<template functional>` is not valid SFC syntax and has no analogue), so the descriptor-level check is **both sound and complete** for the cases the original rule covers — zero false positives (it matches an exact `functional` attribute key on the template block only; the word appearing as element content or another attribute's value is never flagged). The diagnostic span is the opening `<template ...>` tag exactly.

## Tests / gates

- 7 inline `#[cfg(test)]` cases (valid + invalid, including: combined with `lang`, unrelated attrs, `functional` as content/attr-value, plain-template input no-op, presence of a `<script>` block) — all green.
- `cargo test -p vize_patina` (1657 passed) and `cargo test -p vize_carton` (i18n touched) green.
- i18n en/ja/zh rows added to the growable `i18n_supplemental_extra.rs` (pinned `i18n_supplemental.rs` untouched at 347).
- Preset membership snapshot regenerated (appears once, under `opt_in`); `npm/vize/src/types/rules.ts` updated in sorted position.
- Source-length guard respected: no edited file grew past 350; `engine.rs` and `rule.rs` untouched.
- Formatted with the repo's pinned `rustfmt +1.95.0 --edition 2024`.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/1714"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a>

### DIFF
--- a/crates/vize_carton/src/i18n_supplemental_extra.rs
+++ b/crates/vize_carton/src/i18n_supplemental_extra.rs
@@ -174,4 +174,23 @@ static ENTRIES: &[(&str, &str, &str, &str)] = &[
         "オブジェクトを1つにまとめてください。例: :class=\"[{ a }, { b }]\" は :class=\"{ a, b }\" になります。",
         "请将这些对象合并为一个，例如 :class=\"[{ a }, { b }]\" 改为 :class=\"{ a, b }\"。",
     ),
+    // vue/no-deprecated-functional-template
+    (
+        "vue/no-deprecated-functional-template.description",
+        "Disallow the `functional` attribute on the SFC `<template>`",
+        "SFCの`<template>`への`functional`属性を禁止する",
+        "禁止在SFC的`<template>`上使用`functional`属性",
+    ),
+    (
+        "vue/no-deprecated-functional-template.message",
+        "the `functional` attribute on `<template>` (functional SFC templates) was removed in Vue 3",
+        "`<template>`の`functional`属性（関数型SFCテンプレート）はVue 3で削除されました",
+        "`<template>`上的`functional`属性（函数式SFC模板）已在Vue 3中移除",
+    ),
+    (
+        "vue/no-deprecated-functional-template.help",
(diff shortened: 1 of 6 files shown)


ubugeeei:

PR #1715, opened 2026-06-13
feat(patina): add script/require-explicit-emits

ubugeeei:
Implements `script/require-explicit-emits` from #1629 (M4 strongly-recommended). It is roughly the inverse of `script/no-unused-emit-declarations`: instead of flagging declared-but-unemitted events, it flags **emitted-but-undeclared** events.

## What it does

Reports a string-literal emit whose event name is absent from the component's declared emits. Declarations are read from:

- `<script setup>` `defineEmits([...])`, `defineEmits({ ... })`, and the type forms `defineEmits<{ change: [id] }>()` / `defineEmits<{ (e: 'change', id): void }>()`.
- The Options API `emits: [...]` / `emits: { ... }` (resolved via the shared `emits_source` helper, extended here to also return the array form).

Emit call sites are the captured `defineEmits` binding (`const emit = defineEmits(...); emit('x')`, any binding name) and `this.$emit('x')`.

```ts
const emit = defineEmits(['change'])
emit('change') // ok
emit('input')  // reported: 'input' is not declared
```

## Sound subset (vs. upstream `vue/require-explicit-emits`)

patina's markup-rules and script-rules are **separate passes that cannot see each other**, so a `script/*` rule cannot observe template `@evt` / `v-on:evt` usage — which is half of what the upstream rule cross-references. Rather than ship an unsound approximation, this rule restricts itself to what is decidable from the script alone, with **zero/minimal false positives**:

- **Reports only when a declaration exists *and* is fully known.** With no `defineEmits`/`emits` declaration at all, nothing is reported (the emits may be declared elsewhere, or intentionally omitted) — flagging would be a false positive. This is the explicit guidance in the rule's #1629 note.
- **A declaration that cannot be fully enumerated is treated as unknown** (no reports): array/object spread (`defineEmits([...names])`), a computed object key (`{ [k]: null }`), or a bare type reference (`defineEmits<Emits>()`, whose named type isn't visible to the rule).
- **Only string-literal emit names are checked**; a dynamic name (`emit(name)` / `this.$emit(name)`) is skipped.

The template half of upstream (warn on a template-used event that is undeclared) is intentionally **out of scope** here; it would need a template-aware pass.

## Implementation notes

- Rule: `crates/vize_patina/src/rules/script/props_emits/require_explicit_emits.rs` (trait impl + emit-call walking/reporting), with declaration-resolution logic split into a sibling `require_explicit_emits/declared.rs` to keep both files under the 350-line source-length guard.
- Modeled after `no_unused_emit_declarations.rs` and `return_in_emits_validator.rs`; reuses/extends `props_emits/emits_source.rs`.
- Registered with the other props/emits script rules (`OPINIONATED_SCRIPT_PRESETS`, same convention as `require-default-prop` / `require-prop-types`); opt-in by name like its siblings.
- Inline diagnostic messages (no i18n table changes), matching the props/emits batch (#1690).
- 25 inline tests (valid + invalid across all declaration/emit forms, plus the soundness suppression cases). Preset membership snapshot and `npm/vize/src/types/rules.ts` regenerated.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/1715"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a>

### DIFF
--- a/crates/vize_patina/src/linter/script_rules/registry/names.rs
+++ b/crates/vize_patina/src/linter/script_rules/registry/names.rs
@@ -67,6 +67,7 @@ pub(crate) const RULE_REQUIRE_PROP_TYPES: &str = "script/require-prop-types";
 pub(crate) const RULE_NO_RESERVED_PROPS: &str = "script/no-reserved-props";
 pub(crate) const RULE_NO_UNUSED_EMIT_DECLARATIONS: &str = "script/no-unused-emit-declarations";
 pub(crate) const RULE_RETURN_IN_EMITS_VALIDATOR: &str = "script/return-in-emits-validator";
+pub(crate) const RULE_REQUIRE_EXPLICIT_EMITS: &str = "script/require-explicit-emits";
 
 pub(in crate::linter::script_rules) const ALL_BUILTIN_SCRIPT_RULE_NAMES: &[&str] = &[
     RULE_NO_OPTIONS_API,
@@ -122,6 +123,7 @@ pub(in crate::linter::script_rules) const ALL_BUILTIN_SCRIPT_RULE_NAMES: &[&str]
     RULE_NO_RESERVED_PROPS,
     RULE_NO_UNUSED_EMIT_DECLARATIONS,
     RULE_RETURN_IN_EMITS_VALIDATOR,
+    RULE_REQUIRE_EXPLICIT_EMITS,
 ];
 
 #[cfg(test)]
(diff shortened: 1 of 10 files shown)


ubugeeei:

PR #1716, opened 2026-06-13
feat(patina): add vue/no-unused-refs

ubugeeei:
Implements `vue/no-unused-refs` from #1629 (M2): report template refs (`ref="x"`) that are never referenced in `<script>`.

## Scope — sound conservative subset

This rule fundamentally needs to correlate the **template** (`ref="x"` attributes) with the **script** (a `ref` named `x`, `useTemplateRef('x')`, `$refs.x`, …). Patina's markup pass and script pass are separate, so the full ESLint semantics — which build a binding/reference graph over the script AST — cannot be expressed directly. Rather than ship an unsound rule that false-positives on every indirectly-used ref, this is the **soundest conservative subset** (zero false positives by construction):

- It runs as a template `Rule` and, via the shared SFC descriptor, correlates each static `ref="x"` against the **raw `<script>` / `<script setup>` source text**.
- A ref is considered **used** whenever its name appears anywhere in a script block as a whole identifier/string token. That single token test soundly covers every real usage form, because each one contains the literal token in the script:
  - a `<script setup>` variable named `x`;
  - the string key in `useTemplateRef('x')` / `useTemplateRef("x")`;
  - the member access in `this.$refs.x` / `$refs.x` / `$refs['x']`;
  - destructuring `const { x } = this.$refs`;
  - `defineExpose({ x })`, a `computed`/`watch` over `x.value`, etc.
- The rule therefore only ever reports a ref whose name is **textually absent from every script block**, so it never flags a ref that is reachable indirectly.

To stay sound where the text test cannot resolve a target, the rule reports nothing for the file when:

- there is no `<script>` / `<script setup>` block to correlate against;
- a script uses `$refs` **opaquely** — a computed access `$refs[expr]`, optional chaining `$refs?.x`, or `$refs` passed / destructured / iterated as a whole object — since any ref could then be reached without its name appearing literally;
- the **template** itself mentions `$refs` (e.g. `{{ $refs.x.value }}`), since cross-pass resolution of template `$refs` access is unreliable.

Dynamic ref bindings (`:ref` / `v-bind:ref`) are skipped: their target is an expression, not a named template ref.

This direction (report template refs unused **in script**) matches the ESLint rule's intent; the conservative bailouts mirror its `hasUnknown` guards.

## Preset

`opinionated` (and `nuxt`), per the issue's assignment for `vue/no-unused-refs`.

## Tests

18 inline tests cover: unused refs in `<script setup>` and Options API; valid refs via matching variable, `useTemplateRef` with a differing variable name, `this.$refs.x`, `$refs` destructuring, Options-API `setup()` return; dynamic `:ref` / `v-bind:ref` skipping; the no-script and opaque-`$refs` bailouts; template `$refs` usage; nested refs under `v-if` / `v-for`; multiple mixed refs; and substring non-matching (`inputElement` does not satisfy `ref="inputEl"`). A diagnostic snapshot is included. `cargo test -p vize_patina` passes (1668 tests); rustfmt (pinned 1.95.0 / edition 2024) and clippy are clean; the preset membership snapshot and `npm/vize/src/types/rules.ts` are regenerated.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/1716"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a>


ubugeeei:

PR #1977, opened 2026-06-18
fix(check): register transitive imports for default runs

### LINKED ISSUE #1964: bug(check): imports to files outside tsconfig include are typed as 'any' (not loaded transitively) → false-positive TS7031 under strict
## Summary

`vize check` only loads the files matched by a tsconfig's `include`/`files` into the type-check program. When a checked file **imports** a module that resolves to a file **outside** that `include` set, Vize types the import as `any` instead of pulling the target into the program transitively (which is what `tsc` does). Under `strict`, this surfaces as **false-positive `TS7031` (implicitly `any`)** diagnostics on code that `tsc`/`vue-tsc` accept.

Module resolution itself succeeds (no `TS2307`) — the module is found, but its **types are dropped** because the source file isn't in the include set.

This is a big improvement over 0.236.0 (the `../`-glob include bug #1937 is now fixed — thanks!), but this program-construction gap remains and produces false positives in Nuxt's `tsconfig.node.json` context.

## Environment

- `vize` 0.238.0
- `@typescript/native-preview` 7.0.0-dev.20260618.1
- Node v24, macOS

## Minimal reproduction (self-contained, no Nuxt)

```
repro/
  tsconfig.json          # include: ["inside/**/*.ts"]   <-- only inside/
  inside/
    use.ts               # imports from ../outside/lib
  outside/
    lib.ts               # NOT in include
```

`outside/lib.ts`:

```ts
export const ITEMS = [{ code: 'en', name: 'English' }, { code: 'ru', name: 'Russian' }]
```

`inside/use.ts`:

```ts
import { ITEMS } from '../outside/lib'
export const r = ITEMS.map(({ code, name }) => `${code}:${name}`)
```

`tsconfig.json`:

```json
{
  "compilerOptions": { "strict": true, "noEmit": true, "module": "ESNext", "moduleResolution": "Bundler" },
  "include": ["inside/**/*.ts"]
}
```

```sh
vize check          # uses tsconfig.json
tsc -p tsconfig.json
```

## Actual

```
$ vize check
inside/use.ts
  error:2:31 [TS7031] Binding element 'code' implicitly has an 'any' type.
  error:2:37 [TS7031] Binding element 'name' implicitly has an 'any' type.
✗ 2 error(s)
```

`ITEMS` is treated as `any[]` because `outside/lib.ts` is not in the `include` set.

## Expected

```
$ tsc -p tsconfig.json
(exit 0, no errors)
```

`tsc` loads `outside/lib.ts` transitively via the import, so `ITEMS` keeps its type `{ code: string; name: string }[]` and the destructured params are typed. `vize check` should do the same: follow imports into the program even when the target file is outside `include`/`files`.

## Impact — Nuxt

Nuxt's generated `.nuxt/tsconfig.node.json` includes the build/config files (`nuxt.config.ts`, layer `nuxt.config.ts`) but **not** `shared/**`. `nuxt.config.ts` does:

```ts
import { SUPPORTED_LOCALE_METADATA } from './shared/utils/locale'
// ...
locales: SUPPORTED_LOCALE_METADATA.map(({ code, file, name, languageTag }) => ({ /* ... */ }))
```

`vize check --tsconfig .nuxt/tsconfig.node.json` reports:

```
nuxt.config.ts  TS7031 Binding element 'code'/'file'/'name'/'languageTag' implicitly has an 'any' type
```

A probe (`const _p: number = SUPPORTED_LOCALE_METADATA`) produces **no** `TS2322` — confirming Vize sees the import as `any`, not as the array type. `nuxi typecheck` (vue-tsc) reports zero errors on the same file.

On a real Nuxt 4 app, checking the four generated configs (`app`/`server`/`shared`/`node`) on a clean tree now gives **0 errors for app/server/shared** (2600+ files — parity with vue-tsc 🎉) and only `node` is red, entirely due to this transitive-import gap.

## Secondary observation (separate root cause — not minimized)

Same `.nuxt/tsconfig.node.json` run also reports:

```
layers/auth/nuxt.config.ts:15  TS2353 Object literal may only specify known properties,
                               and 'routeRules' does not exist in type 'InputConfig<NuxtConfig, ConfigLayerMeta>'.
```

`routeRules` is a valid Nuxt/Nitro config key (declaration-merged onto `NuxtConfig`). `vue-tsc` accepts it. This looks like Vize not merging module/`declare module` augmentations of `NuxtConfig`, which is a different issue from the transitive-import one above — flagging it here for visibility; happy to open a separate issue with a focused repro if useful.

Prains:
## Summary
- Register reachable local source imports for default tsconfig-based `vize check` runs.
- Keep diagnostics limited to original tsconfig inputs while using transitive files for type resolution.
- Add regression coverage for outside-include imports in default runs.

Fixes #1964

## Test Plan
- [x] cargo test -p vize --tests
- [x] git diff --check

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/1977"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784382254&installation_id=136613492&pr_number=1977&repository=ubugeeei-prod%2Fvize&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fvize%2Fpull%2F1977&signature=42bf70118eb8fb2945ac29ae1124fabe0b6a7a41e9acc72df0ae4ff5982c953f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/crates/vize/src/commands/check/imports.rs
+++ b/crates/vize/src/commands/check/imports.rs
@@ -1,9 +1,9 @@
-//! Transitive resolution of source imports for explicit `vize check` subsets.
+//! Transitive resolution of source imports for `vize check` virtual projects.
 //!
-//! `vize check src/App.vue` registers only the requested files in the virtual
-//! project, so a sibling import like `./types` or `~/components/Foo.vue` cannot
-//! see the real types and can surface false `TS2307`. This module walks the
-//! reachable graph and returns on-disk source files to register.
+//! A check run may intentionally report diagnostics for only a subset of
+//! sources, but imported local sources still need to be registered so
+//! cross-file types resolve like tsc/vue-tsc. This module walks the reachable
+//! graph and returns on-disk source files to register.
 
 use std::path::{Path, PathBuf};
 
--- a/crates/vize/src/commands/check/runner/diagnostics.rs
+++ b/crates/vize/src/commands/check/runner/diagnostics.rs
@@ -18,25 +18,20 @@ pub(super) fn emit_json_output(json_output: JsonOutput) {
     }
 }
 
-/// Whether a registered file's diagnostics should be reported. For an explicit
-/// subset (`reported` is `Some`), only the requested files are reported; ambient
-/// and transitively-registered files exist only to resolve cross-file types.
-/// Project-level diagnostics (anchored to a tsconfig or the project root, not
-/// a source file) describe the whole check and are always reported.
+/// Whether a registered file's diagnostics should be reported. Only listed
+/// source files are reported; ambient and transitively-registered files exist
+/// only to resolve cross-file types. Project-level diagnostics (anchored to a
+/// tsconfig or the project root, not a source file) describe the whole check and
+/// are always reported.
 pub(super) fn is_reported(
-    reported: &Option<FxHashSet<PathBuf>>,
+    reported: &FxHashSet<PathBuf>,
     path: &Path,
     canonical_paths: &mut CanonicalPathCache,
 ) -> bool {
-    match reported {
-        None => true,
-        Some(set) => {
-            if !is_source_path(path) {
-                return true;
-            }
-            set.contains(&canonical_paths.canonicalize(path))
-        }
+    if !is_source_path(path) {
+        return true;
     }
+    reported.contains(&canonical_paths.canonicalize(path))
 }
 
 fn is_source_path(path: &Path) -> bool {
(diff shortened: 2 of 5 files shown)


ubugeeei:
Thank you, it have some conflicts, and I'll take it

Issue #2836, opened 2026-07-12
chore(project): track GPT-5.6 Sol Ultrra review findings


ubugeeei:
## Goal

Track the implementation, correctness, performance, release-safety, and DX findings from the full-repository **GPT-5.6 Sol Ultrra Review**.

Principles for this backlog:

- one behavioral fix per PR;
- preserve or improve measured performance;
- add the smallest executable regression contract;
- use conventional titles;
- require all relevant GitHub Actions checks to finish successfully;
- use auto-merge only after required and relevant optional checks are green.

## P0 / release and package integrity

- [x] #2815 — merged in #2838
- [x] #2846 — exact-SHA gates merged in #2852
- [x] #2853 — local release publication guard merged in #2854
- [x] #2856 — release fuzz evidence fails on crashes, merged in #2857
- [x] #2860 — benchmark dispatch rejects invalid comparison ranges, merged in #2861
- [x] #2866 — deterministic Fuzz evidence identity merged in #2867
- [x] #2874 — rerun-safe crate packaging/frontier dry-run merged in #2877
- [x] #2879 — exact crate-version registry state handling merged in #2882
- [x] #2873 — explicit editor-registry publication merged in #2885
- [x] #2818 — hardened exact-SHA release preflight merged in #2893, #2896–#2899

## P1 / correctness and data safety

- [x] #2827 — merged in #2844
- [x] #2828 — merged in #2840
- [x] #2841 — merged in #2842
- [x] #2821 — merged in #2837
- [x] #2823 — source-relative build outputs merged in #2850
- [x] #2822 — merged in #2843
- [x] #2824 — merged in #2847
- [x] #2825 — merged in #2851
- [x] #2826 — bounded per-call native batch pools merged in #2845
- [x] #2816 — merged in #2839
- [x] #2817 — merged in #2835
- [x] #2829 — WASM option contract aligned in #2858
- [x] #2832 — Vite/Musea config instances isolated in #2862
- [x] #2849 — output write failures propagated in #2850
- [x] #2848 — literal File/Directory/Glob classification merged in #2875
- [x] #2864 — isolated, diagnostic Corsa LSP smoke merged in #2865
- [x] #2871 — general build glob semantics merged in #2881
- [x] #2872 — distinct symlink-parent sources preserved in #2883
- [x] #2886 — unused slot-outlet prop hoists removed in #2887

## P1 / CI and contributor DX

- [x] #2813 — merged in #2814
- [x] #2819 — merged in #2820
[...]

PR #2907, opened 2026-07-12
fix(lsp): refresh dependent component diagnostics

### LINKED ISSUE #2905: fix(lsp): refresh diagnostics for changed component importers
## Problem

When an open Vue component changes, diagnostics are only republished for that component. Open Vue files that import and use it in their templates keep stale type errors (or miss newly introduced errors) until the importer itself is edited.

## Root cause

`textDocument/didChange` updates the changed component's virtual document and publishes diagnostics only for that URI. Corsa receives the dependency update, but no diagnostic request is made for open importers.

## Acceptance criteria

- After a Vue dependency changes, open direct Vue importers are diagnosed again without editing them.
- Revalidation is scoped to actual open importers rather than every open document.
- Relative imports in both `<script>` and `<script setup>` are handled.
- An LSP integration test covers a child prop type change surfacing in the parent template.

Parent review: #2836

ubugeeei:
## Summary

- maintain an O(1)-lookup reverse dependency index for open Vue importers
- republish importer diagnostics after the changed Corsa virtual document is synchronized
- preserve unsaved Vue and script dependency buffers while Corsa rebuilds an importer project
- cover relative imports from both normal script and script setup blocks
- add an end-to-end regression for a child prop type change surfacing in its parent

## Validation

- `cargo test -p vize_canon corsa_bridge::vue_document_tests --lib`
- `cargo test -p vize_maestro server::importers --lib`
- `cargo build -p vize`
- `cargo clippy -p vize_maestro --lib -- -D warnings`
- `node --test tests/tooling/lsp-typecheck-template.test.ts`

Closes #2905
Parent review: #2836

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved LSP type-checking so diagnostics refresh correctly for already-open parent components when imported Vue files change (including prop type mismatches).
  * Diagnostics publishing now updates only the affected documents and refreshes diagnostics for relevant open Vue importers.
* **New Features**
  * Virtual Vue project building now prefers overlay (in-memory/unsaved) dependency contents during dependency scanning.
* **Tests**
  * Added an LSP regression test for refreshed parent diagnostics after child prop type changes.
  * Added a unit test ensuring overlay-provided types are preferred over on-disk sources.

### DIFF
--- a/crates/vize_canon/src/corsa_bridge/vue_dependencies.rs
+++ b/crates/vize_canon/src/corsa_bridge/vue_dependencies.rs
@@ -4,7 +4,7 @@ use std::collections::VecDeque;
 use std::path::{Path, PathBuf};
 
 use oxc_span::SourceType;
-use vize_carton::{FxHashSet, String, cstr};
+use vize_carton::{FxHashMap, FxHashSet, String, cstr};
 
 use super::bridge::normalize_document_uri;
 use super::vue_dependency_specifiers::collect_relative_ts_specifiers;
@@ -22,6 +22,7 @@ pub(super) fn collect_dependency_documents(
     host: &GeneratedVueDocument,
     options: CorsaVueVirtualDocumentOptions,
     rewriter: &ImportRewriter,
+    overlays: &FxHashMap<PathBuf, &str>,
 ) {
     let mut visited_vue = FxHashSet::<PathBuf>::default();
     visited_vue.insert(host.source_path.clone());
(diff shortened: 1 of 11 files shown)


ubugeeei:
Addressed the review findings in `02271a930`:

- diagnostic refresh now owns virtual-document synchronization;
- open Vue importers are maintained in a reverse dependency index, so change lookup is O(1) instead of reparsing every open file;
- extensionless Vue/TS/TSX/JS/JSX imports and index modules are resolved by the index;
- importer Corsa projects prefer open dependency buffers, preventing disk content from overwriting unsaved component changes;
- the integration assertion now matches stable observable behavior rather than exact compiler wording.

Validated with the focused Corsa project tests, importer index tests, clippy, build, and the real two-file LSP change flow.

ubugeeei:
@coderabbitai review

PR #3006, opened 2026-07-17
fix(lsp): harden package import resolution

ubugeeei:
Follow-up to #3005 (already merged), addressing CodeRabbit review findings on the package-import resolution added there. The original branch was merged and deleted before these could land, so they ship here off `main`.

Three correctness/security fixes in the open-Vue reverse-dependency importer:

**Exact `.` and `..` relative specifiers.** Node/TS treat bare `.` and `..` as relative imports (resolving to `./index` and `../index`), but the dispatch only matched `./` and `../` prefixes, so they fell through to package resolution and failed to resolve.

```rust
if specifier == "."
    || specifier == ".."
    || specifier.starts_with("./")
    || specifier.starts_with("../")
{
    return resolve_relative_import(importer_dir, specifier);
}
```

**Reject export targets that escape the package root.** `package_root.join(target)` accepted absolute paths and `../` components, so malformed dependency metadata could associate an import with files outside the package. `resolve_package_target` now rejects parent/root/prefix components before joining (while still allowing bare-relative targets like `main: "index.js"`).

```rust
let relative = Path::new(target.strip_prefix("./").unwrap_or(target));
if relative.components().any(|component| {
    matches!(component, Component::ParentDir | Component::RootDir | Component::Prefix(_))
}) {
    return None;
}
resolve_package_candidate(package_root.join(relative))
```

**Array export fallback at the file level.** `conditional_export_target` collapsed an array to its first string, so a first entry that failed to resolve to an existing file blocked later valid fallbacks. `package_export_target` now returns all candidate targets in priority order and resolution tries each until one resolves.

Added regressions for `.`/`..` specifiers, an escaping export target, and a broken-first/valid-second array; existing direct-call assertions updated for the new `Vec` return.

Validated with `cargo check -p vize_maestro` and `cargo fmt`. Full clippy/test run is left to CI (the sandbox disk can't build the test binary).

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/3006"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved import discovery for directory-style specifiers such as `"."` and `".."`.
  * Added support for resolving multiple package export targets, including fallback to the first valid option.
  * Prevented package imports from resolving outside the package directory through malformed paths.
  * Improved handling of conditional and array-based package exports.

### DIFF
--- a/crates/vize_maestro/src/server/importers.rs
+++ b/crates/vize_maestro/src/server/importers.rs
@@ -144,7 +144,11 @@ fn resolve_import(importer_dir: &Path, specifier: &str) -> Option<PathBuf> {
     let specifier = specifier
         .split_once(['?', '#'])
         .map_or(specifier, |(path, _)| path);
-    if specifier.starts_with("./") || specifier.starts_with("../") {
+    if specifier == "."
+        || specifier == ".."
+        || specifier.starts_with("./")
+        || specifier.starts_with("../")
+    {
         return resolve_relative_import(importer_dir, specifier);
     }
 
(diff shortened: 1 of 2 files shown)


ubugeeei:
Superseded by #3007, which merged the same fixes with full local and CI validation. The remaining conditional-export edge case from review will be handled in a separate focused follow-up.

PR #3035, opened 2026-07-18
test(fixtures): harden atelier matrix runner and assertions

ubugeeei:
Follow-up to #3034 (already merged) addressing CodeRabbit review feedback on the compiler fixture matrix.

- Stop leaking the compiler temp directory when a fixture is missing. `runTool` now allocates the `mkdtempSync` output dir only when the fixture exists, so the early `missing-fixture` return (which bypasses the `try/finally` cleanup) no longer strands a temp directory.

```js
const fixtureExists = existsSync(cwd);
const compilerOutputDir =
  tool === "compiler" && !args.dryRun && fixtureExists
    ? mkdtempSync(join(tmpdir(), "vize-fixture-compiler-"))
    : null;
// ...
if (!fixtureExists) return { ...base, status: "missing-fixture" };
```

- Decouple the report test from shell-quote formatting: the compiler command is now asserted flag-by-flag, matching the `<compiler-output>` value without requiring literal surrounding quotes.
- Assert the exact deterministic SHA-256 for the fixed synthetic artifact instead of only matching the hex format, so ordering or digest-input regressions are caught.

Not addressed: the suggestion to reconcile artifact `filename` values against the matched input paths. The compiler sets `filename` to the file basename (`path.file_name()` in `crates/vize/src/commands/build/runner/compile.rs`), and real projects have many colliding basenames across directories, while output files are written common-root-relative rather than cwd-relative. Reconciling on those identities would inject false failures into the real matrix, so the existing count + envelope checks are retained.

Validation: `node --test --test-concurrency=1 tests/tooling/fixture-tool-matrix-compiler.test.ts tests/tooling/fixture-tool-matrix-report.test.ts` (11 pass), plus oxlint and oxfmt over the three changed files.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/3035"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a>

### DIFF
--- a/tests/tooling/fixture-tool-matrix-compiler.test.ts
+++ b/tests/tooling/fixture-tool-matrix-compiler.test.ts
@@ -86,7 +86,10 @@ process.stdout.write("Built: vize-tool-matrix-test.vue\\n");`,
           },
           { inputFileCount: 1, outputFileCount: 1, errorCount: 0, warningCount: 1 },
         );
-        assert.match(raw.compilerArtifacts.sha256, /^[0-9a-f]{64}$/);
+        assert.equal(
+          raw.compilerArtifacts.sha256,
+          "472e9ac54a4318345412c8fab1f81aa472684be9d7c9791adea885dc5ef19a3f",
+        );
         assert.equal("parsed" in raw, false);
       } finally {
         fs.rmSync(outputDir, { recursive: true, force: true });
--- a/tests/tooling/fixture-tool-matrix-report.test.ts
+++ b/tests/tooling/fixture-tool-matrix-report.test.ts
@@ -62,10 +62,13 @@ test("fixture tool matrix emits read-only commands with machine-readable diagnos
     const runs = Object.fromEntries(
       report.projects[0].runs.map((entry: { tool: string }) => [entry.tool, entry]),
     ) as Record<string, { command: string }>;
-    assert.match(
-      runs.compiler.command,
-      /build .*--format json --output '<compiler-output>' --template-syntax quirks --continue-on-error --no-config/,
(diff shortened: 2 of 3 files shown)


ubugeeei:
Closing as obsolete. The merge conflict here resolves to an empty diff: every change on this branch was independently and more completely implemented by the already-merged #3036 ("fix(fixtures): reconcile atelier artifacts"):

- The `missing-fixture` temp-dir leak fix in `tools/fixtures/tool-matrix-run.mjs` is byte-identical to #3036 (it auto-merges to nothing).
- The flag-by-flag `--output` report assertions in `fixture-tool-matrix-report.test.ts` are functionally equivalent to #3036's.
- The exact pinned compiler SHA-256 in `fixture-tool-matrix-compiler.test.ts` is superseded by #3036's `706ebeac…`; this branch's `472e9ac…` is stale because #3036 renamed the synthetic artifact to `vize-tool-matrix-test.json`, changing the digest input.

Additionally, #3036 implemented the artifact reconciliation that this PR had deliberately rejected. Resolving the conflict in favor of `main` (the only correct choice, since keeping this branch's versions would revert #3036's merged reconciliation and reintroduce a stale SHA) yields a tree identical to `main`. There is nothing left to merge. Reopen if I've missed something.

PR #3199, opened 2026-07-21
test(canon): regenerate ecosystem-products canon snapshot for static prop value checks

ubugeeei:
## Why

The static-attribute prop type-checking change from #3197 was merged into `main`, but its `vize check --show-virtual-ts` snapshot for the `ecosystem-products` fixture was never regenerated. As a result the `vue-parity` job now fails on `main` with a snapshot mismatch:

```
✖ vize check does not crash and snapshot matches
  Error: Snapshot mismatch: tests/snapshots/check/__snapshots__/ecosystem-products-check.snap
  Run with UPDATE_SNAPSHOTS=1 to update.
```

The mismatch is limited to the `virtualTs` fields of the 5 fixture files that pass static attribute values to child components.

## What

Regenerates `ecosystem-products-check.snap` with the current binary. The change is purely additive: the new per-prop assertions vize now emits for static attribute values, e.g.

```ts
type __NButton_1_prop_type = __VizePropValue<__NButton_Props_1, 'type'>;
// @vize-map: prop -> 2262:2276
const __vize_prop_check_1_type: __NButton_1_prop_type = "primary";
void __vize_prop_check_1_type;
```

`diagnostics` per file, `errorCount` (2), `warningCount` (0), and `fileCount` (6) are all unchanged, confirming the newly-checked static values type-check cleanly and introduce no new false positives. Only the recorded generated TypeScript grew.

Follow-up to #3197.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/3199"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a>


ubugeeei:
Superseded by #3200, which merged the byte-identical snapshot refresh (this branch now has a zero diff against main).

PR #3212, opened 2026-07-22
fix(maestro): keep prop inlay hints out of literal text

ubugeeei:
## Summary

Fixes the reported inlay-hint bug: in a binding like

```
:class="[
  'tag',
  `tag--size-${size}`,
  `tag--state-${state}`,
]"
```

the `#props.` hint appeared **twice per line** — once on the real `${size}` reference and once on the static template text `tag--size-`, because `find_prop_usages_in_expr` anchored on every word-boundary occurrence of the prop name in the raw expression text.

## Fix

Hint matching now runs only over the expression's **executable code regions**, computed by a small scanner (`inlay_hint/expr_regions.rs`): string literals, template-literal text, and comments are excluded, while `${...}` interpolations re-enter code — through nested template literals and with balanced object braces, so `${fn({ size })}` still hints and `}` inside an interpolation doesn't end it early. All existing guards (word boundaries, `props.`-prefix, member access, event-name segments) still apply within regions.

## Tests

- Scanner unit tests: template text vs interpolations, strings and both comment kinds, nested templates, object braces inside interpolations, unterminated literals.
- End-to-end regression pinning the reported component shape to **exactly one hint per real reference** (5 hints: the `aria-disabled` comparison plus four interpolations), each asserted to sit on a code reference rather than literal text.
- `vize_maestro` suite green (559), fmt clean, clippy at the pre-change baseline.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/3212"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved inlay hints in template expressions by excluding static template text, quoted strings, and comments.
  * Ensured hints are shown only for valid executable code, including nested interpolations and complex expressions.
  * Added coverage for unterminated and nested template literals to improve hint accuracy.


ubugeeei:
@codesmith-bot 

fix conflict

Issue #3223, opened 2026-07-22
roadmap(patina): win the eslint-plugin-vue parity scorecard


ubugeeei:
Tracks the linter product workstream (peer of the #3131 tree and #3222).

## Outcome

`patina` (via oxlint-plugin-vize) beats `eslint-plugin-vue` on rule coverage, finding parity, template-awareness, fix safety, and speed — machine-checked.

## Scorecard dimensions

- [ ] Rule-by-rule differential: pin `eslint-plugin-vue` + `eslint` versions; checked-in `rule-map.json` (eslint ruleId → patina ruleId | `intentional-divergence:<reason>` | `unimplemented:<issue>`); run `vue/recommended` over the real-project corpus; assert identical finding locations modulo documented divergences; unmapped findings tracked in a `vize.fixtureLintDivergence` ledger mirroring `tools/fixtures/typecheck-divergence.mjs`.
- [ ] Template-awareness audit: systematic sweep for script-only rules that must see template usage (the #3209/#3210/#3211 defect class), with cross-block fixtures for every rule that references bindings.
- [ ] Fix safety: property tests that autofixes preserve parse validity and never change runtime semantics on the ecosystem corpus (fix → recompile → output-diff clean).
- [ ] Lint/format agreement: `vize fmt` output never triggers patina formatting-category findings (corpus-wide property, shared with the glyph suite).
- [ ] Performance: keep the existing bench lanes green while closing coverage gaps; publish a rules-per-ms comparison vs eslint-plugin-vue.

## Delivery

Small conventional PRs, one dimension slice per PR, auto-merge after green.

ubugeeei:
Audited dimension 1. **The rule-map half is built and pinned; the corpus-differential half is not.**

## Built and passing

`tests/_fixtures/patina-eslint-vue-rule-map.json` is exactly the checked-in map this dimension asks for, with the upstream pin the dimension requires:

```json
"upstream": { "package": "eslint-plugin-vue", "version": "10.9.2", "ruleCount": 252 }
"summary": { "mapped": 120, "unimplemented": 132 }
```

[...]

PR #3265, opened 2026-07-23
fix(glyph): preserve whitespace-significant and template-literal content

### LINKED ISSUE #3247: fix(glyph): idempotence violation on multiline template literals
Found by the corpus idempotence property (#3238, part of #3223). `vize fmt --write` re-indents the inner lines of multiline template literals relative to the surrounding code, and does so again on every pass, so `fmt(fmt(x)) != fmt(x)` and the literal's string contents (which are semantically significant) drift further on each run. The same root cause also trips the parse-preservation property: the template literal appears inside a template interpolation, and its quasis differ between `x` and `fmt(x)`.

Affected corpus files (all copies of the same shadcn-vue ChartStyle component):
- `tests/_fixtures/_git/shadcn-vue/apps/v4/registry/bases/reka/ui/chart/ChartStyle.vue`
- `tests/_fixtures/_git/shadcn-vue/apps/v4/registry/new-york-v4/ui/chart/ChartStyle.vue`
- `tests/_fixtures/_git/shadcn-vue/apps/v4/styles/reka-{luma,lyra,maia,mira,nova,rhea,sera,vega}/ui/chart/ChartStyle.vue`

Repro: run `vize fmt --write --no-config` twice over any of these files; pass 2 reports the file reformatted again, with growing indentation inside the backtick template:

```
fmt1 >       ${prefix} [data-chart=${id}] {
fmt2 >             ${prefix} [data-chart=${id}] {
```

Template literal content must be treated as opaque text (bytes preserved). Skip-listed in `tests/_fixtures/glyph-corpus-known-violations.json` under idempotence and parse-preservation until fixed.

### LINKED ISSUE #3249: fix(glyph): parse-preservation violation on split closing tags
Found by the corpus parse-preservation property (#3238, part of #3223). When a closing tag is split across lines (the Prettier-style `</pre\n>` trick used to keep whitespace out of `<pre>` content), `vize fmt` reprints the closing tag AND leaves the original `>` behind as a stray text node, changing the rendered output.

Affected corpus file: `tests/_fixtures/_git/vue-router/packages/playground-file-based/src/pages/test-params/query.vue`

Before:
```html
<pre>
page (number): {{ route.params.page }} ({{ typeof route.params.page }})</pre
  >
```

After `vize fmt --write --no-config`:
```html
  <pre>
page (number): {{ route.params.page }} ({{ typeof route.params.page }})</pre>
  >
```

The second line now renders a literal `>`. Caught as `template > <pre>[4]: element pre -> text ">"` by the @vue/compiler-sfc structural comparison. Skip-listed in `tests/_fixtures/glyph-corpus-known-violations.json` under parse-preservation until fixed.

### LINKED ISSUE #3250: fix(glyph): parse-preservation violation on whitespace-only textarea content
Found by the corpus parse-preservation property (#3238, part of #3223). `vize fmt` drops whitespace-only text content inside `<textarea>`, collapsing `<textarea ...>\n    </textarea>` to `<textarea ...></textarea>`. Textarea content is its initial value, so this changes the DOM (HTML parsing only strips one leading newline; the remaining indentation was part of the value).

Affected corpus files:
- `tests/_fixtures/_git/vux/src/demos/Issue624.vue`
- `tests/_fixtures/_git/mint-ui/packages/field/src/field.vue`

Caught as `<textarea>[0]: child 0 disappeared (#text)` by the @vue/compiler-sfc structural comparison (text under `pre`/`textarea` is compared byte-for-byte). Skip-listed in `tests/_fixtures/glyph-corpus-known-violations.json` under parse-preservation until fixed.

ubugeeei:
Three corpus-property violations from #3223 all trace to the template formatter changing bytes that are semantically significant. Each is fixed surgically and its skip-list entry removed so the property now enforces the case.

### #3250 — parse-preservation: whitespace-only `<textarea>` dropped
Whitespace-only `<textarea>` content is the element's initial value. The immediate-empty-close collapse ran before the whitespace-significant check and dropped it, changing the DOM. Guard the collapse:

```rust
} else if !is_void
    && !is_whitespace_significant_element(&tag_name, &sorted_attrs)
    && let Some(closing_end_pos) =
        self.parse_immediate_empty_closing_tag(source, end_pos, &tag_name)
{
```

Whitespace-significant elements now fall through to the verbatim-copy branch.

### #3249 — parse-preservation: split closing tag left a stray `>`
A closing tag split across lines (`</pre
  >`, the Prettier trick that keeps the trailing newline out of `<pre>` content) was consumed as a bare `</pre>` of fixed length, leaving `  >` behind as a stray text node that then rendered a literal `>`. Scan to the real `>`:

```rust
pos = memchr::memchr(b'>', &source[close_start..])
    .map_or(len, |off| close_start + off + 1);
```

### #3247 — idempotence + parse-preservation: multiline template literals re-indented
Multiline interpolations were re-indented line by line, which prepends indentation to the raw content of a multiline template literal (the shadcn-vue `ChartStyle` shape). That both corrupts the value and is not a fixed point: the indentation grows on every pass. The fix tracks template-literal quasi state and emits quasi lines verbatim while still indenting the surrounding code:

```rust
for (idx, line) in trimmed.lines().enumerate() {
    if quasi_line_starts.get(idx).copied().unwrap_or(false) {
        out.push_str(line); // inside a quasi: preserve bytes exactly
    } else {
        self.write_indent_string(&mut out, depth + 1);
        out.push_str(line.trim_end_matches('\r'));
    }
    out.push_str(self.newline_str());
}
```

The state machine handles nested `${ … }` interpolations (which oxc canonicalizes each pass) and skips ordinary strings/escapes so their braces and backticks don't confuse it.

### Verification
`cargo test -p vize_glyph` passes (108 tests). Added regression tests covering each case, including the real nested `ChartStyle` shape (idempotent, outer quasi preserved verbatim).

Closes #3247
Closes #3249
Closes #3250

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/3265"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Vue template formatting for whitespace-sensitive elements such as `<pre>`, `<textarea>`, and `v-pre`.
  * Prevented stray characters from appearing around split closing tags.
  * Preserved whitespace and template-literal content within interpolations.
  * Improved formatter consistency so repeated formatting produces stable results.

* **Tests**
  * Added regression coverage for interpolation formatting, whitespace preservation, and closing-tag handling.


ubugeeei:
Splitting into one focused PR per issue (#3247, #3249, #3250) as requested. Superseded shortly.

Issue #2971, opened 2026-07-17
test(editor): establish real-world typecheck and LSP production gates


ubugeeei:
## Status (revalidated against `origin/main` `865280f86`, 2026-08-03)

**25 of 25 items satisfied.** Every discrete gap from the prior audit is closed and completed; the enforcing gates shipped in v0.322.0.

## Goal

Prove that Vize's typechecker, LSP, and editor integrations deliver a fast, robust, type-safe authoring experience on real Vue projects. This deepens the completed feature roadmap in #676: feature presence is not enough unless clean projects stay clean, intentional defects are found at the authored span, fixes clear stale state, and large workspaces remain responsive.

## Non-negotiable verification model

- Treat pinned repositories in `tests/_fixtures/_git` as read-only upstream inputs.
- Copy only the exercised files into an isolated work directory before applying a test patch; never commit changes to an upstream gitlink.
- Record both sides of every oracle:
  - clean baseline: no unexpected Vize-only diagnostic;
  - broken patch: the exact expected diagnostic code, severity, file, authored range, and stable message fragment appears;
  - repaired patch: the diagnostic disappears without restarting the LSP, and dependent files refresh when relevant.
- Compare typechecker results with the configured `vue-tsc` baseline where the fixture registry enables `typecheckPerformance`; classify every divergence as a Vize false positive, a Vize false negative, or an intentional documented difference.
- Assert editor-visible results, not only internal virtual TypeScript: diagnostics, completion must-include and must-exclude sets, hover type/documentation, definition/reference/rename ranges, code actions, and protocol version ordering.
- Keep every implementation PR narrow, use a conventional title, enable auto-merge, and wait for all required checks and review state before calling it complete.

## Fixture tiers

- [x] Tier S: `create-vue` generated templates for fast clean/broken/repaired patch oracles.
- [x] Tier M: Vue Router and Pinia library workspaces for package exports, project references, declarations, generics, and cross-package navigation.
- [x] Tier UI: Element Plus or Reka UI for slots, emitted events, generic components, directives, and declaration-heavy component APIs.
- [x] Tier framework: Nuxt and VitePress for generated ambient types, aliases, auto-imports, virtual modules, and no-tsconfig/best-effort behavior.
- [x] Tier legacy: a pinned Vue 2 / Vue 2.7 Options API and class-component project.
- [x] Tier L: Misskey and Vue Vben Admin for large monorepo discovery, dependency invalidation, cancellation, memory, and latency budgets.

[...]

PR #3324, opened 2026-07-28
test(editor): pin authored-range mapping for hover, definition, references, rename, semantic tokens, inlay hints, and document features

ubugeeei:
Delivers audit item 8 of #2971:

> Verify hover, definition, references, rename, code actions, semantic tokens, inlay hints, and document features map to authored Vue ranges.

Editor features answer by querying a generated virtual TypeScript document and mapping the result back to the authored SFC. A mapping regression therefore shows up as a whole-line or whole-column shift — exactly the failure mode that survives a `length > 0` or `.some(...)` assertion. Half of this PR is the audit of what was already pinned; the other half closes the gaps it found.

## Audit: exact-authored-range coverage before this PR

Every row was verified against a real spawned `vize lsp` process (`LspSession`), not a mocked transport.

| Feature | Exact authored range pinned before? | Strongest prior assertion | This PR |
| --- | --- | --- | --- |
| **hover** | **No** — ~10 real-server tests, all regex-on-markdown; `Hover.range` never asserted anywhere | `tests/tooling/lsp-editor-production-gates.test.ts:155` — 6 regex matches on markdown | Pins the exact type resolved *through* the authored template position; **documented skip** for the range gap found |
| **definition** | Yes | `tests/tooling/lsp-definition-navigation.test.ts:74,136` — start + end, plus literals | Adds template→script and event-handler→script over a pinned real-world SFC |
| **references** | Partial — one exhaustive test; elsewhere existential `.some(...)`, one matches **line only** | `tests/tooling/lsp-definition-navigation.test.ts:169-217` | Must-include authored ranges + single-URI must-exclude; **documented skip** for the defect found |
| **rename** / **prepareRename** | Yes | `tests/tooling/lsp-smoke.test.ts:426-458`, `tests/tooling/lsp-rename-guards.test.ts:98-176` | Full edit-list `deepEqual` in both directions + `Object.keys(changes)` must-exclude |
| **code actions** | **Yes — already exact**, left untouched | `tests/tooling/lsp-smoke.test.ts:466-566` pins the diagnostic range, the fix edit range, and the suppression insert range | Not changed |
| **semantic tokens** (`/full`, `/range`) | **No** — only `data.length > 0` and `length % 5 === 0`; a systematic shift passed every test | `tests/tooling/lsp-workspace-semantic.test.ts:310-342`, `tests/tooling/lsp-smoke.test.ts:258-305` | Decodes the 5-tuple delta stream through the **advertised legend** to absolute authored positions; asserts the authored text each token covers; asserts `/range` re-bases deltas without shifting absolute positions |
| **inlay hints** | Partial — labels/kind/tooltip exact, `position.line` only; **`position.character` never asserted** | `tests/tooling/lsp-inlay-codelens.test.ts:46-105` | Full `deepEqual` including exact `position.character` at the authored declaration end |
| **document symbols** | Partial — names, kinds, `selectionRange.end.character` literals; never a whole range object | `tests/tooling/lsp-document-features.test.ts:104-291` | Full `deepEqual` of `range` **and** `selectionRange` for all three blocks |
| **folding ranges** | **No** — only `startLine < endLine`, never concrete lines | `tests/tooling/lsp-document-features.test.ts:293-348` | Exact `startLine`/`endLine` |
| **document highlight** | Yes — strongest prior coverage in the repo | `tests/tooling/lsp-document-highlight.test.ts:112-209` | Adds a real-fixture case that doubles as the differential isolating the references defect |
| **document links** | Yes — start and end | `tests/tooling/lsp-document-link.test.ts:118-139,222-243` | Not changed |

## Gaps found (documented skips, not aspirational assertions)

Both skips carry the **real acceptance criterion in the test body**, so unskipping is the whole fix-verification step. Both were confirmed to fail today by temporarily unskipping them against a freshly built binary.

**1. `textDocument/hover` never populates the optional `Hover.range`.** The response contains only `contents`, so an editor cannot underline the hovered authored span. Observed: `hover.range === undefined`.

**2. `textDocument/references` returns unmapped virtual-TS coordinates for script-side hits.** Querying `visitCount` (declared line 3, used line 4, used in the template line 9) answers:

```
[ 3:6-16, 4:6-16, 5:31-41, 9:18-28 ]
```

The declaration `3:6-16` and the template use `9:18-28` are correct. The two extra entries are the *virtual* positions of the two script occurrences, one line below authored — and `5:31-41` cannot exist at all, because authored line 5 is the 9-character `</script>`. The authored script use at `4:31-41` is therefore **absent** from the result. `textDocument/rename` and `textDocument/documentHighlight` map the identical occurrence set correctly on the same fixture, so this is references-only and not a shared mapping bug. The unskipped assertion in this PR pins that every returned range must address text that exists in the file, which is what catches it.

## Evidence

Every expected value was derived from probe runs against a freshly built binary (`cargo build --profile ci -p vize --features legacy`), never guessed, and each probe was repeated in-session to confirm the responses are deterministic. Representative raw responses over the fixture:

```
definition@template-doubled  {"range":{"start":{"line":4,"character":6},"end":{"line":4,"character":13}}}
rename@script-visitCount     [3:6-16, 4:31-41, 9:18-28] newText "visits"
documentHighlight            [{kind:3, 3:6-16}, {kind:2, 4:31-41}, {kind:2, 9:18-28}]
semanticTokens/full data     [3,19,3,12,512, 1,16,8,12,512, 4,9,7,8,0, 1,10,6,11,0, 0,8,10,8,0, 0,10,2,21,0]
inlayHint                    [{": Ref<number>", 3:16}, {": ComputedRef<number>", 4:13}]
foldingRange                 [template 7-14, script setup 0-5]
```

The semantic-token stream decodes through the advertised legend to `ref` (3:19, function/defaultLibrary), `computed` (4:16), `doubled` (8:9, variable — the template interpolation), `@click` (9:10, event), `visitCount` (9:18, variable), `++` (9:28, operator). The test asserts the authored substring each token covers, so a column shift can no longer pass.

## Notes

- Uses the pinned `create-vue` fixture, already hydrated by the parity gate, so no hydration-list change is needed. The upstream `App.vue` is SHA-256 pinned: if upstream moves, the suite fails loudly instead of silently shifting every expected range.
- Single `vize lsp` session and one workspace for all ten sub-tests (~0.2s), so the gate cost is negligible.
- No existing assertion or snapshot was weakened; the suites nearest the change (`lsp-document-features`, `lsp-inlay-codelens`, `lsp-document-highlight`, `lsp-definition-navigation`, 18 tests) still pass unchanged.

## Remaining gaps for follow-up (not addressed here, to keep this PR small)

- **completion** — no `textEdit` / replace-range assertion exists anywhere, so a mis-anchored insertion would go undetected.
- **formatting / rangeFormatting** — only the `null` path is ever asserted through a real server. `range_formatting` validates `params.range` then formats the *entire* document; the capability is deliberately unadvertised.
- **codeLens** — `tests/tooling/lsp-inlay-codelens.test.ts:156` characterizes `range.start.line === declLine + 1`, the same one-line offset seen in the references defect.
- **documentSymbol on single-line blocks** — the `<style scoped></style>` block range collapses to zero width because open and close tags share a line. Pinned as observed with a comment.
- **Not implemented** (consistent with unadvertised capabilities, zero tests): `selectionRange` (returns `Method not found`), `signatureHelp`, `typeDefinition`, `implementation`, `documentColor`.

## Test plan

```
cargo build --profile ci -p vize --features legacy

VIZE_TEST_BIN=target/ci/vize VIZE_LSP_BIN=target/ci/vize VIZE_TEST_REQUIRE_TSGO=1 \
  node --test --test-concurrency=1 \
    snapshots/check/create-vue-editor-range-oracle.ts \
    snapshots/check/create-vue-patch-oracle.ts \
    snapshots/check/create-vue-generated-template-oracle.ts
# 14 pass, 2 documented skips, 0 fail
```

Part of #2971

## Summary by CodeRabbit

* **Tests**
  * Added an expanded fixture and compiler validation test script (`test:check:fixtures`).
  * Added LSP test utilities to construct and assert authored ranges/positions, and to validate semantic token decoding against the legend.
  * Declared a new snapshot test for the editor-range “oracle” assertions to ensure exact authored Vue ranges are covered.


ubugeeei:
@codesmith-bot fix CI

Issue #3294, opened 2026-07-28
fix(patina): template parse recovery silently disables analysis-dependent rules


ubugeeei:
Found while hydrating the pinned `vue2-elm` fixture in the vue-parity lane (#3293): the glyph corpus `lint-agreement` property reports `vue/no-unused-vars: 0 -> 1` for `src/page/shop/shop.vue`, and the growth is a pristine-side false negative rather than a formatter defect.

The pristine file has `version="1.1"class="specs_cancel"` at 272:101, which `vize lint` surfaces as a recovered `parser/template` diagnostic ("Missing whitespace between attributes; inferred a new attribute boundary"). On that recovered parse croquis semantic analysis is unavailable, so every rule guarded by `has_analysis()` returns without reporting, including `vue/no-unused-vars` (`crates/vize_patina/src/rules/vue/no_unused_vars.rs:98`).

`vize fmt --write` normalizes the attribute boundary. The formatted file parses cleanly, analysis becomes available, and the genuinely unused `index` in `<transition v-for="(item, index) in showMoveDot">` is reported. Nothing is lost or introduced by formatting; the pristine lint simply never ran the rule.

Repro (fresh `cargo build --profile ci -p vize --features legacy`):

```
mkdir -p /tmp/repro/src/page/shop
cp tests/_fixtures/_git/vue2-elm/src/page/shop/shop.vue /tmp/repro/src/page/shop/
cd /tmp/repro
vize lint 'src/**/*.vue' --format json --preset ecosystem --no-config   # 1 parser/template, 0 vue/no-unused-vars
vize fmt  'src/**/*.vue' --write --no-config
vize lint 'src/**/*.vue' --format json --preset ecosystem --no-config   # 0 parser/template, 1 vue/no-unused-vars at 934:21
```

Analysis-dependent rules should still run when the template parser recovers (a recovered parse is exactly the shape where a silent skip is least acceptable), so the pristine tree and the formatted tree report the same rules.

Until then the disagreement is pinned rule-scoped in `tests/_fixtures/glyph-corpus-known-violations.json` so the corpus property stays strict everywhere else.

PR #3327, opened 2026-07-28
fix(canon): keep type checking a file whose template parse recovered (#3323)

### LINKED ISSUE #3323: fix(canon): a recovered template parse error silences every script type diagnostic in the file
Found while gating audit item "parse-error recovery keeps independent diagnostics while suppressing only dependent noise" (#2971).

## Minimal reproduction

Two SFCs with **identical** script blocks. The only difference is one missing space between two attributes — a defect the template parser recovers from with an inferred attribute boundary.

`src/Clean.vue`:
```vue
<template>
  <div>{{ label }}</div>
</template>

<script setup lang="ts">
const label: string = 'hi'
const wrong: number = 'also not a number'
</script>
```

`src/Recovered.vue` — same script, template has `id="a"class="b"`:
```vue
<template>
  <div id="a"class="b">{{ label }}</div>
</template>

<script setup lang="ts">
const label: string = 'hi'
const wrong: number = 'not a number'
</script>
```

`vize check 'src/**/*.vue' --format json` under a strict tsconfig:

```json
{
  "files": [
    { "file": "src/Clean.vue",     "diagnostics": ["error:7:7 [TS2322] Type 'string' is not assignable to type 'number'."] },
    { "file": "src/Recovered.vue", "diagnostics": ["error:2:14 Template parse error: Missing whitespace between attributes; inferred a new attribute boundary."] }
  ]
}
```

Restoring the single space (same file, same script) brings the type error straight back:

```
Fixed.vue (whitespace restored): ["error:7:7 [TS2322] Type 'string' is not assignable to type 'number'."]
```

## Expected

`Recovered.vue` reports **both**: the recovered template parse error *and* the independent script type error at 7:7. The script block's correctness does not depend on an attribute boundary in the template.

## Actual

The script diagnostic disappears entirely. One missing space silences all type checking for that file — a false negative that hides real type errors behind an unrelated cosmetic defect, and it scales: any file with a recovered parse defect stops being type-checked while `vize check` still exits reporting only the parse error.

## Relationship to #3294

Same failure shape, different subsystem. #3294 was the linter: a recovered parse classified as fatal skipped croquis analysis, so `has_analysis()`-guarded rules silently reported nothing. That fix (#3310, #3311) introduced `CompilerError::is_recovered_parse()` and the shared `RECOVERED_PARSE_CODES` list in `vize_relief`, keyed the linter off it, and pinned the classification against the parser's recovery-message table.

The type-check path was not converted. The same classification should decide whether virtual-TS generation proceeds: a recovered parse yields a complete tree, so the script block can still be generated and checked, and only genuinely template-dependent diagnostics should be suppressed.

## Verified against

`cargo build --profile ci -p vize --features legacy` at `d9b7fe19d`.

ubugeeei:
## Summary

A single missing space between two attributes silenced every type diagnostic in the file. Two SFCs with identical script blocks:

```
src/Clean.vue      -> error:7:7 [TS2322] Type 'string' is not assignable to type 'number'.
src/Recovered.vue  -> error:2:14 Template parse error: Missing whitespace between attributes; ...
```

`Recovered.vue` differs only by `id="a"class="b"` in its template. Restoring the space brings the TS2322 straight back, so the type error was real and simply not reported.

`generate_vue_virtual_ts` treated every non-`ExtendPoint` parse error as hard: it dropped the template AST and returned `invalid_sfc_fallback_virtual_ts()`, so the script block was never generated and never checked. But the parser documents a concrete recovery for thirteen codes (`vize_relief::errors::recovery::RECOVERED_PARSE_CODES`) and still yields a complete tree for them.

This is the same false-negative shape #3294 fixed in the linter, keyed off the same shared classification introduced by #3310/#3311 — the type-check path just had not been converted. Now:

- a code with a documented recovery reports its parse diagnostic **and** keeps the real virtual TS;
- script parse errors (no AST at all) still abort to the stub, tracked separately from template diagnostics;
- an unrecovered template error still aborts exactly as before.

After the fix, `Recovered.vue` reports both diagnostics — the parse defect at 2:14 and the type error at 7:7.

## Verification

- New `vize_canon` regression tests: a recovered attribute boundary keeps the real virtual TS with both script bindings generated (fails before the fix), and an unrecovered parse still collapses to the fallback stub (passes on both sides, pinning the preserved behavior).
- End-to-end with a fresh `--profile ci --features legacy` binary on the reproduction above: 2 errors before (parse-only on the recovered file), 3 after (parse **and** type on the recovered file), with the clean twin unchanged.
- `cargo test -p vize_canon`: 19 suites green, 0 failures. fmt clean.

Closes #3323

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of Vue files with recoverable template syntax issues, preserving generated TypeScript output when possible.
  * Script parsing failures now correctly trigger safe fallback behavior.
  * Unrecoverable template syntax errors continue to use the fallback output.

* **Tests**
  * Added coverage for recovered template parsing and unrecoverable syntax scenarios.

### DIFF
--- a/crates/vize_canon/src/batch/virtual_project/vue_codegen.rs
+++ b/crates/vize_canon/src/batch/virtual_project/vue_codegen.rs
@@ -110,12 +110,22 @@ pub(super) fn generate_vue_virtual_ts(
         .as_ref()
         .map(|template| template.loc.start as u32)
         .unwrap_or(0);
+    // Script parse errors leave no AST, so they always abort to the fallback
+    // stub. Template diagnostics do not: they are counted separately below and
+    // only *hard* ones abort.
+    let script_hard_error = !diagnostics.is_empty();
     // Track whether the template produced any *hard* parse error. Only hard
-    // errors abort codegen and collapse the file to the fallback stub;
-    // recovery-level diagnostics (`ErrorCode::ExtendPoint`, pushed by the HTML
-    // tree-construction recovery path for self-closing rewrites, fostered
-    // elements, auto-closed `<p>`, etc.) describe repairs the parser already
-    // applied and must keep the real virtual TS (#1065/#1090 regression).
+    // errors abort codegen and collapse the file to the fallback stub.
+    // Recovery-level diagnostics keep the real virtual TS:
+    //   - `ErrorCode::ExtendPoint`, pushed by the HTML tree-construction
+    //     recovery path for self-closing rewrites, fostered elements,
+    //     auto-closed `<p>`, etc. (#1065/#1090 regression);
+    //   - every code in `RECOVERED_PARSE_CODES`, for which the parser
+    //     documents a concrete recovery and still yields a complete tree.
+    //     Treating those as hard meant a single missing space between
+    //     attributes collapsed the file to the stub and silenced every script
+    //     type diagnostic in it (#3323) — the same false-negative shape #3294
+    //     fixed in the linter, keyed off the same shared classification.
     let mut template_hard_error = false;
     let template_ast = descriptor.template.as_ref().and_then(|template| {
         profile!("canon.template.parse", {
@@ -132,7 +142,11 @@ pub(super) fn generate_vue_virtual_ts(
                 if error.code.is_recovery() {
                     continue;
                 }
-                template_hard_error = true;
+                // A documented recovery still yields a complete tree, so the
+                // defect is reported without suppressing the rest of the file.
+                if !error.code.has_documented_parse_recovery() {
+                    template_hard_error = true;
+                }
                 let start = error
                     .loc
                     .as_ref()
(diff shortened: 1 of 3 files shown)


ubugeeei:
@codesmith-bot fix CI

ubugeeei:
Pushed fixes for the two red required checks, plus the snapshot refresh the fix implies.

**`test-scripts` (source length)** — the new test module added one line to `crates/vize_canon/src/batch/virtual_project/tests.rs`, which sits at exactly 1337 lines and is already over the 350-line budget, so it may not grow. (The file has no blank lines between its `use`, `mod`, and `fn` blocks — it was already squeezed to fit.) The tests moved to `crates/vize_canon/tests/recovered_parse_codegen.rs` as an integration test, using the public `VirtualProject` API like the sibling suites there, and the `mod` line came back out. Same two tests, same assertions, both still fail before the fix.

**`vue-parity` (vue2-elm snapshot)** — expected, and the delta is the fix working. Refreshed with `UPDATE_SNAPSHOTS=1`, then diffed as a **multiset of (file, diagnostic) pairs** rather than by line, since the snapshot reflows:

```
errorCount 273 -> 313
added 40, removed 0
files receiving new diagnostics: { src/page/shop/shop.vue: 40 }
codes: TS2339 x20, TS1308 x8, TS2349 x3, TS2551 x3, TS2588 x2, TS2304 x2, TS2307 x2
fileCount 55 -> 55, warningCount 0 -> 0
```

Every added diagnostic lands on `src/page/shop/shop.vue` — the one file in the corpus with a recovered template parse error (`version="1.1"class="specs_cancel"`, the same defect #3294 was about). That file was previously collapsed to the fallback virtual TS, so **none** of its type errors were reported; it is now checked like the other 54. The new diagnostics are the same families the rest of the corpus already reports (`vuex`/`better-scroll` TS2307, `__VizeThis` TS2339, TS2349, TS1308), so nothing novel appeared — one silently-unchecked file simply started being checked.

Nothing was removed, so no previously-reported diagnostic was lost.

PR #3328, opened 2026-07-28
fix(canon): resolve extensionless .vue imports through path mappings

### LINKED ISSUE #3300: fix(canon): extensionless webpack-style .vue imports stay TS2307 despite paths mapping and ambient shims
Found while pinning the vue2-elm check suite (#3263, PR #3293).

## Minimal reproduction

vue2-elm (pinned at `tests/_fixtures/_git/vue2-elm`) imports SFCs webpack-style without extensions through path aliases, e.g. `import Head from 'src/components/header/head'` where the target is `head.vue`, with `paths` mapping configured and ambient `declare module` shims present.

## Expected behavior

Resolution follows the alias to the `.vue` file (or at minimum consumes the ambient module shim), so no module-not-found diagnostic.

## Actual behavior

TS2307 "Cannot find module" — 125 of the 353 pinned diagnostics in `tests/snapshots/check/__snapshots__/vue2-elm-check.snap` are this class; ambient `declare module` shims are not consumed for these specifiers.

The snapshot (PR #3293) pins the current surface; fixing this shrinks it and refreshing the snapshot documents the win.

ubugeeei:
## Root cause

Not the ambient shims — the fix is entirely in path-mapping resolution.

The pinned vue2-elm fixture contains **no `.d.ts` at all**, and canon's own
`__vize_vue_modules.d.ts` is an inert marker comment (canon deliberately
materializes real `.vue.ts` modules instead of a `declare module "*.vue"`
wildcard). A `*.vue` wildcard shim could not match an extensionless specifier
anyway, so option (b) from the issue is a dead end.

What actually happens: `remap_paths` re-anchors each tsconfig `paths` target
into two candidates — the virtual-mirror copy and the real-tree fallback:

```json
"paths": { "src/*": ["./src/*", "../../../src/*"] }
```

For `import Head from 'src/components/header/head'` neither resolves. The
mirror holds `src/components/header/head.vue.ts`; the real tree holds
`head.vue`. TypeScript *appends* extensions to the mapped specifier rather
than replacing them, so it probes `head.ts`/`head.tsx`/`head.d.ts`/`head.js`
and `head/index.*` — never `head.vue.ts`. Result: TS2307.

(Note `baseUrl` is stripped from the generated tsconfig, so `paths` is the
only hook for bare specifiers.)

## Fix

Each relative alias target gains a third candidate, `./<target>.vue.ts`,
appended **after every ordinary candidate** so it can only turn a previously
failing resolution into a success — it can never shadow a real-tree `.ts` or
a later user-declared target. Targets that already name a concrete module
(`.ts`, `.tsx`, `.vue`, `.json`, `.js`, ...) and absolute targets are skipped.

```json
"paths": { "src/*": ["./src/*", "../../../src/*", "./src/*.vue.ts"] }
```

The import specifiers themselves are untouched, so no virtual offsets shift
and diagnostic source-mapping is unaffected.

New logic lives in `crates/vize_canon/src/batch/virtual_project/tsconfig_gen/vue_alias.rs`
(135 lines); `tsconfig_gen.rs` shrinks 454 → 446 lines.

## Snapshot delta

`tests/snapshots/check/__snapshots__/vue2-elm-check.snap`: **353 → 273
diagnostics, 80 removed, 0 added.** `fileCount` (55) and `warningCount` (0)
unchanged.

Verified by a per-file multiset diff of the JSON reports (not by reading the
line diff, which reflows because JSON array entries shift). **All 80 removed
diagnostics are TS2307 of the extensionless-alias class:**

| count | specifier |
| ----: | --------- |
| 41 | `src/components/header/head` |
| 20 | `src/components/common/alertTip` |
| 10 | `src/components/common/loading` |
| 4 | `src/components/footer/footGuide` |
| 2 | `src/components/common/shoplist` |
| 1 | `src/components/common/computeTime` |
| 1 | `src/components/common/ratingStar` |
| 1 | `src/components/common/buyCart` |

The remaining 45 TS2307 are genuinely uninstalled vendor packages (`vuex` 38,
`better-scroll` 6, `showdown` 1) and are untouched. No diagnostic of any other
code was removed, and resolving the components introduced no new errors.

## Tests

- New unit tests on the target expansion: wildcard alias, extensionless
  directory alias, concrete-module targets keeping the original pair,
  ordering (vue candidates never precede an ordinary candidate), and
  absolute/non-string passthrough.
- Updated the three existing materialized-tsconfig assertions
  (`materialized_tsconfig_reanchors_paths_into_virtual_mirror`,
  `materialized_tsconfig_reanchors_extended_paths_from_declaring_config_dir`,
  `materialized_project_keeps_generated_graphql_schema_on_real_path`).
- `cargo test -p vize_canon --features legacy`: 604 passed, 0 failed.
- `cargo test --workspace --features legacy`: green except
  `check_nuxt2_false_positive_fixture_matrix`, which fails identically on
  unmodified `main` in this environment (TS2339 `$emit`, unrelated).
- `npm run test:check:fixtures` (with `VIZE_TEST_BIN`/`VIZE_TEST_TSGO` pinned
  to the ci binary): the three failures are all pre-existing on `main` here
  (unhydrated vue-router submodule, vue-tsc manifest lookup, generic-build
  injected-error fixture).
- `cargo fmt --all --check` clean; clippy reports nothing on the touched files.

## Known remaining gap (not this PR)

**Relative** extensionless SFC imports are still unresolved — e.g.
`src/App.vue` does `import svgIcon from './components/common/svg'` for
`svg.vue`. tsgo raises TS2307 for 6 of these inside the virtual project, but
`relative_module_resolves_on_disk` suppresses the diagnostic because the
`.vue` sibling exists on disk, so they never appeared in the snapshot and the
count above is unaffected. The diagnostic is hidden but the module is
genuinely unresolved. Fixing it needs the import rewriter to gain the source
directory on the `.vue` path (`build_vue_registered_file` calls
`rewriter.rewrite`, which has neither roots nor source dir) and shifts
specifier offsets, so it is deliberately left out of this change.

Closes #3300

### DIFF
--- a/crates/vize_canon/src/batch/virtual_project/tests.rs
+++ b/crates/vize_canon/src/batch/virtual_project/tests.rs
@@ -929,11 +929,11 @@ fn materialized_tsconfig_reanchors_paths_into_virtual_mirror() {
         serde_json::from_str(&fs::read_to_string(tsconfig_path).unwrap()).unwrap();
     let paths = value["compilerOptions"]["paths"].as_object().unwrap();
 
-    // Each target gets a mirror candidate (relative to the virtual tsconfig
-    // in `node_modules/.vize/canon`) first, then the real-tree fallback.
+    // Mirror candidate first, then the real-tree fallback, then a trailing
+    // `.vue.ts` mirror candidate for extensionless SFC aliases (#3300).
     assert_eq!(
         paths["@/*"],
-        serde_json::json!(["./src/*", "../../../src/*"])
+        serde_json::json!(["./src/*", "../../../src/*", "./src/*.vue.ts"])
     );
     assert_eq!(
         paths["#shared"],
@@ -984,14 +984,14 @@ fn materialized_tsconfig_reanchors_extended_paths_from_declaring_config_dir() {
         serde_json::from_str(&fs::read_to_string(tsconfig_path).unwrap()).unwrap();
     let paths = value["compilerOptions"]["paths"].as_object().unwrap();
 
-    assert_eq!(
-        paths["~/*"],
-        serde_json::json!(["./app/*", "../../../app/*"])
-    );
-    assert_eq!(
-        paths["#imports"],
-        serde_json::json!(["./.nuxt/imports", "../../../.nuxt/imports"])
-    );
+    let app = ["./app/*", "../../../app/*", "./app/*.vue.ts"];
+    let imports = [
+        "./.nuxt/imports",
+        "../../../.nuxt/imports",
+        "./.nuxt/imports.vue.ts",
+    ];
+    assert_eq!(paths["~/*"], serde_json::json!(app));
+    assert_eq!(paths["#imports"], serde_json::json!(imports));
 
     let _ = fs::remove_dir_all(&case_dir);
 }
--- a/crates/vize_canon/src/batch/virtual_project/tests/graphql_generated.rs
+++ b/crates/vize_canon/src/batch/virtual_project/tests/graphql_generated.rs
@@ -108,7 +108,7 @@ expectQuestion(question)
         serde_json::from_str(&fs::read_to_string(tsconfig_path).unwrap()).unwrap();
     assert_eq!(
         value["compilerOptions"]["paths"]["~/*"],
-        serde_json::json!(["./*", "../../../*"])
+        serde_json::json!(["./*", "../../../*", "./*.vue.ts"])
     );
 
     let _ = fs::remove_dir_all(&case_dir);
--- a/crates/vize_canon/src/batch/virtual_project/tsconfig_gen.rs
+++ b/crates/vize_canon/src/batch/virtual_project/tsconfig_gen.rs
@@ -1,4 +1,5 @@
 mod native_options;
+mod vue_alias;
 
 use std::path::{Path, PathBuf};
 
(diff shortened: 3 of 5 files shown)


ubugeeei:
Note on the red `pr-benchmark-budget` check (non-required, did not block the merge): its `Compile SFC: 1.086x (+8.57%)` is measurement noise, not a real regression.

Evidence:

- **Causally implausible**: this PR only changes generated-tsconfig `paths` expansion, which `vize build` never touches — the benchmark runs `vize build . --format stats --threads 1`.
- **Reran the check**: identical to three decimals (`1.086x (+8.57%)`) in 7s, i.e. it re-evaluates the stored artifact rather than re-measuring, so the rerun cannot distinguish noise from signal.
- **Measured locally instead**, ci-opt binaries at `519cec599` (base) vs `2e4fc75f6` (this PR merged), on the same 1000-file `bench/__in__` corpus, single-threaded, 2 warmups discarded, base/head interleaved to cancel drift:

  ```
  base median 0.1401s (n=10)
  head median 0.1400s (n=10)
  ratio 0.9998x (-0.02%)
  ```

  Both binaries report `✓ 1000 files compiled`, so the corpus is genuinely processed.

The other four PRs open at the same time passed this check, so the noise is not systemic — but a lane that can produce an 8.6% false positive at a 5% threshold, and whose rerun replays the stored number, will keep costing triage time. Worth either raising the threshold to cover observed variance or making the rerun actually re-measure.

PR #3331, opened 2026-07-28
fix(vapor): replace multi-step _next(base, N) with _nthChild or chained _next

### LINKED ISSUE #3330: vapor: emitted `_next(node, N)` multi-step navigation crashes on vue 3.6.0-rc.2 (`next()` advances one sibling when not hydrating)
## Summary

Vize's vapor codegen emits `_next(node, 2)` / `_next(node, 3)` calls that appear to expect multi-step sibling advancement. In vue 3.6.0-rc.2's runtime, `next(node, logicalIndex)` honors `logicalIndex` only during hydration; otherwise it advances exactly one sibling (`_next(node) === node.nextSibling`). Navigation therefore lands on the wrong node, and a chained `_child()` dereferences `null`:

```
TypeError: Cannot read properties of null (reading 'firstChild')
```

## Versions

- vize / @vizejs/vite-plugin: **0.291.0**
- vue: **3.6.0-rc.2**
- vite-plus: 0.2.6 (`vite` = `@voidzero-dev/vite-plus-core`; note `@vizejs/vite-plugin` imports `parseSync` from `vite`, which upstream vite does not provide)

## Reproduction

Minimal repro: https://github.com/mymx2/vize/tree/repro/vapor-next-nav-crash/examples/vapor-next-nav-repro

```sh
pnpm install
pnpm dev
# open http://localhost:5173 → console shows the TypeError above (3x)
```

The component (any template where the compiler must navigate ≥2 siblings to reach a bound node seems affected):

```vue
<script setup lang="ts">
import { ref } from 'vue';
const logo = ref('data:image/svg+xml,<svg xmlns="http://www.w3.org/2000/svg"/>');
</script>

<template>
  <div>
    <svg role="presentation"><use href="#x" /></svg>
    <h2><span>Title</span></h2>
    <p><span>Sub</span></p>
    <ul>
      <li><a href="https://example.com"><img :src="logo" alt="" /><span>A</span></a></li>
      <li><a href="https://example.org"><img :src="logo" alt="" /><span>B</span></a></li>
    </ul>
  </div>
</template>
```

## Emitted output (`vize.config.ts` → `compiler: { vapor: true }`)

```js
const t0 = _template("<div><svg ...>...</svg><h2>...</h2><p>...</p><ul>...</ul></div>", true);
function render(_ctx, ...) {
  const n1 = t0();
  const n0 = _next(_child(n1), 3);  // meant: svg → ul (3 siblings over)
  const n2 = _child(n0);            // actual: svg → h2 (1 step), then span, then text node
  const n4 = _child(n2);
  const n5 = _child(n4);            // _child(textNode) === null → TypeError on .firstChild
  ...
  _setProp(n5, "src", _ctx.logo);  // intended target: first <img>
}
```

## Runtime contract (vue 3.6.0-rc.2, `@vue/runtime-vapor`)

```js
  if (isHydrating) return locateChildByLogicalIndex(node.parentNode, logicalIndex);
  return _next(node); // exactly one sibling otherwise
}
```

## Expected vs Actual

- **Expected**: navigation reaches the intended nodes (`ul → li → a → img`) and bindings apply; component mounts.
- **Actual**: navigation is off by `N - 1` steps; chained `_child()` hits `null` and the component crashes on mount.

## Notes

- Workaround: avoid global `compiler.vapor: true`. With per-file detection (the `vapor` attribute on `<script setup>`), affected components can stay on vdom while the rest of the app uses vapor.
- Related crash family: #1377 (same `null.firstChild` symptom from misaligned vapor navigation, marked fixed).

## Validations

- [x] Reproduced on the exact versions listed above with the linked minimal example.
- [x] Checked existing issues (#1377, #3073) — both closed, but this navigation-contract variant still reproduces on 0.291.0.

mymx2:
vize's vapor codegen emitted `_next(base, offset)` for child refs with an offset > 1, but the vue 3.6.0-rc.2 runtime's `next()` helper only advances one sibling outside hydration; the second argument is the hydration-only logical index. That caused `null.firstChild` errors when navigating past multiple static siblings in vapor mode.

This change:
- Emits `_nthChild(parent, offset)` for absolute child positions (offset > 1), matching compiler-vapor's contract.
- Chains single-step `_next()` calls instead of passing an offset > 1.

Fixes #3330

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/3331"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787855670&installation_model_id=422833&pr_number=3331&repository=ubugeeei-prod%2Fvize&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fvize%2Fpull%2F3331&signature=fba782a983f712b3ffd1e26b7741482e69415c2dd7d9b9cf669573e3b770db00"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved navigation through deeply nested and adjacent elements.
  * Corrected reference generation for multi-step sibling offsets, ensuring each step is traversed accurately.
  * Preserved optimized handling for direct and single-step navigation.

### DIFF
--- a/crates/vize_atelier_vapor/src/generate/operations/refs.rs
+++ b/crates/vize_atelier_vapor/src/generate/operations/refs.rs
@@ -87,10 +87,19 @@ pub(super) fn generate_child_ref(ctx: &mut GenerateContext, child_ref: &ChildRef
             "const n{} = _child(n{})",
             child_ref.child_id, child_ref.parent_id
         ));
-    } else {
+    } else if child_ref.offset == 1 {
         ctx.use_helper("next");
-        let expr = build_next_chain(cstr!("_child(n{})", child_ref.parent_id), child_ref.offset);
+        let expr = build_next_chain(cstr!("_child(n{})", child_ref.parent_id), 1);
         ctx.push_line_fmt(format_args!("const n{} = {}", child_ref.child_id, expr));
+    } else {
+        // Outside hydration the runtime `next()` advances a single sibling, so
+        // absolute child positions beyond one step must go through `nthChild`
+        // (mirrors vue's compiler-vapor).
+        ctx.use_helper("nthChild");
+        ctx.push_line_fmt(format_args!(
+            "const n{} = _nthChild(n{}, {})",
+            child_ref.child_id, child_ref.parent_id, child_ref.offset
+        ));
     }
 }
 
@@ -104,7 +113,16 @@ pub(super) fn generate_next_ref(ctx: &mut GenerateContext, next_ref: &NextRefIRN
 fn build_next_chain(base: String, offset: usize) -> String {
     if offset == 0 {
         base
+    } else if offset == 1 {
+        cstr!("_next({}, 1)", base)
     } else {
-        cstr!("_next({}, {})", base, offset)
+        // Outside hydration the runtime `next()` only honors single-step
+        // advancement, so multi-step navigation must be a chain of calls
+        // instead of one call carrying the offset.
+        let mut expr = base;
+        for _ in 0..offset {
+            expr = cstr!("_next({})", expr);
+        }
+        expr
     }
 }
--- a/crates/vize_atelier_vapor/src/snapshots/vize_atelier_vapor__tests__compile_dynamic_child_after_multiple_static_siblings.snap
+++ b/crates/vize_atelier_vapor/src/snapshots/vize_atelier_vapor__tests__compile_dynamic_child_after_multiple_static_siblings.snap
@@ -1,12 +1,12 @@
 ---
-source: crates/vize_atelier_vapor/src/lib.rs
+source: crates/vize_atelier_vapor/src/tests.rs
 expression: code.as_str()
 ---
-import { child as _child, next as _next, setClass as _setClass, renderEffect as _renderEffect, template as _template } from 'vue';
+import { child as _child, setClass as _setClass, nthChild as _nthChild, renderEffect as _renderEffect, template as _template } from 'vue';
 const t0 = _template("<div><header>one</header><p>two</p><button>x</button></div>", true)
 export function render(_ctx) {
 const n1 = t0()
-const n0 = _next(_child(n1), 2)
+const n0 = _nthChild(n1, 2)
 _renderEffect(() => _setClass(n0, _ctx.cls))
 return n1
 }


ubugeeei:
Thanks for both the report and the fix — the diagnosis is exactly right, and I confirmed the same contract against the runtime pinned in this repo (`@vue/runtime-vapor@3.6.0-beta.10`), so it is not specific to rc.2.

The `_nthChild` half for `generate_child_ref` matches what `@vue/compiler-vapor` does and looks right to me.

One concern on the other half — the chained `_next` in `build_next_chain`, which `generate_next_ref` still reaches:

```rust
let mut expr = base;
for _ in 0..offset {
    expr = cstr!("_next({})", expr);
}
```

Those chained calls pass no second argument, so during hydration each one takes the `isHydrating` branch with `logicalIndex === undefined`:

```js
  return _next(node)
}
function locateChildByLogicalIndex(parent, logicalIndex) {
  let child = parent.$llc || parent.firstChild
  let fromIndex = child.$idx || 0
  if (logicalIndex < fromIndex) { child = parent.firstChild; fromIndex = 0 }
  while (child) {
    if (fromIndex === logicalIndex) { ... }
    ...
    fromIndex++
  }
  return null
}
```

With `logicalIndex` undefined, `undefined < fromIndex` is false and `fromIndex === undefined` is never true, so the loop runs to the end and returns `null`. Client-side rendering is fixed either way, but a hydrated (SSR) app would hit `null` at that navigation instead.

Upstream's `genChildren` avoids this by never chaining: it uses `next(prev, logicalIndex)` **only** when the target is exactly one sibling on, and `nthChild(from, elementIndex, logicalIndex)` for anything further — the same helper you already used for the child-ref case.

I had #3332 in flight when yours landed in the queue; it takes that route for the `NextRef` side too, which needed `NextRefIRNode` to carry the parent id and the absolute rendered index. Happy for either to be the one that lands — yours is the report and the diagnosis, and if you would rather extend this PR with the `nthChild` form for `generate_next_ref`, I will close mine. I have taken auto-merge off #3332 so it does not race yours.

One more thing worth knowing for whichever lands: the existing snapshot `test_compile_dynamic_child_after_multiple_static_siblings` had **pinned the defect** (`_next(_child(n1), 2)`), so it was asserting output that applies `setClass` to the wrong element at runtime.

ubugeeei (approved):

PR #3332, opened 2026-07-29
fix(vapor): use nthChild for multi-sibling navigation (#3330)

### LINKED ISSUE #3330: vapor: emitted `_next(node, N)` multi-step navigation crashes on vue 3.6.0-rc.2 (`next()` advances one sibling when not hydrating)

ubugeeei:
Reported by @mymx2 in #3330 with a minimal reproduction — thank you, the report was precise enough to confirm the runtime contract directly.

## The defect

The Vue Vapor runtime's `next(node, i)` advances **exactly one** sibling outside hydration. The second argument is an absolute logical index consulted only while hydrating, never a step count:

```js
  return _next(node) // one sibling
}
```

Verified against the runtime pinned in this repo (`@vue/runtime-vapor@3.6.0-beta.10`), so this is not specific to the reporter's 3.6.0-rc.2.

Vize emitted `_next(base, offset)` for **any** offset, treating that argument as a step count. A three-sibling jump landed one sibling over, and the chained `_child()` that followed dereferenced `null` — the reported `TypeError: Cannot read properties of null (reading 'firstChild')`.

## The fix

`@vue/compiler-vapor` picks its navigation helper by distance, and Vize now follows the same rule:

| distance | emitted |
|---|---|
| 0 | `_child(parent)` |
| 1 | `_next(base, index)` — index is the hydration hint |
| ≥ 2 | `_nthChild(parent, index)` — absolute, correct in both modes |

No IR node gained fields: `ChildRefIRNode` already carries a parent id plus an absolute rendered index and already emits `_nthChild(parent, index)`, so a jump of two or more siblings is emitted as a `ChildRef` and `NextRefIRNode` keeps its three public fields (adding fields to it would have been a public API break in a fix release). The transform tracks the previous sibling's rendered index and picks the node kind from index arithmetic instead of a second scan; `NextRefIRNode::offset` now holds the absolute rendered index it passes to `next()` as the hydration hint, matching `ChildRefIRNode::offset`.

## Verification

On the reporter's exact template, before → after:

```diff
-const n0 = _next(_child(n1), 3)   // lands on <h2>; _child chain then hits null
+const n0 = _nthChild(n1, 3)       // lands on <ul>
 const n2 = _child(n0)
 const n4 = _child(n2)
 const n5 = _child(n4)
 const n3 = _next(n2, 1)           // adjacent <li> stays a one-step _next
```

Traced by hand against the template: `n1` = the `<div>`, children `svg(0) h2(1) p(2) ul(3)`, so `_nthChild(n1, 3)` is the `<ul>`, and `n5`/`n7` are the two `` elements the `:src` bindings target.

**An existing snapshot had pinned the defect**: `test_compile_dynamic_child_after_multiple_static_siblings` asserted `_next(_child(n1), 2)` for a two-sibling jump, which at runtime applies `setClass` to the wrong element. Refreshed to `_nthChild(n1, 2)`.

New regression tests (`tests_sibling_navigation.rs`): the reporter's template, a sweep asserting no emitted `_next` ever carries an index above 1 from `_child`, and a test pinning that distances 0 and 1 keep their existing shapes. The first two fail before the fix; the third passes on both sides, pinning preserved behavior.

`cargo test -p vize_atelier_vapor`: 107 passed, 0 failed. fmt and clippy clean.

## Note on hydration

Vize passes its rendered-node index as the hydration hint. Upstream distinguishes `logicalIndex` (which accounts for fragment anchor comments) from `elementIndex`; the two coincide unless fragment anchors sit among the siblings. Hydration with fragment anchors in the sibling list is therefore not covered by this change, and is worth its own issue if it turns up.

Closes #3330

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/3332"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sibling navigation during hydration for dynamic template-backed children.
  * Corrected multi-sibling jumps to use absolute `nthChild` lookups instead of counted `_next` chains.
  * Kept adjacent sibling traversal efficient as a single-step `_next`.

* **Tests**
  * Added regression coverage to confirm correct generated navigation helpers across multiple template shapes, including adjacent and multi-sibling cases.
  * Verified existing behavior for index 0 and index 1 remains unchanged.

### DIFF
--- a/crates/vize_atelier_vapor/src/generate/operations/refs.rs
+++ b/crates/vize_atelier_vapor/src/generate/operations/refs.rs
@@ -100,19 +100,14 @@ pub(super) fn generate_child_ref(ctx: &mut GenerateContext, child_ref: &ChildRef
     ctx.push_line_fmt(format_args!("const n{} = {}", child_ref.child_id, expr));
 }
 
-/// Generate NextRef (`_next` / `_nthChild` helper).
+/// Generate NextRef (`_next` helper).
+///
+/// The node is always the sibling immediately after `prev_id` — jumps of two
+/// or more siblings arrive as a [`ChildRefIRNode`] absolute lookup instead
+/// (#3330) — so this is a single `_next` step carrying the absolute index as
+/// the hydration hint.
 pub(super) fn generate_next_ref(ctx: &mut GenerateContext, next_ref: &NextRefIRNode) {
-    let expr = if next_ref.offset > 1 {
-        ctx.use_helper("nthChild");
-        cstr!("_nthChild(n{}, {})", next_ref.parent_id, next_ref.index)
-    } else {
-        build_navigation(
-            cstr!("n{}", next_ref.prev_id),
-            next_ref.offset,
-            next_ref.index,
-            ctx,
-        )
-    };
+    let expr = build_navigation(cstr!("n{}", next_ref.prev_id), 1, next_ref.offset, ctx);
     ctx.push_line_fmt(format_args!("const n{} = {}", next_ref.child_id, expr));
 }
 
--- a/crates/vize_atelier_vapor/src/generate/operations/refs.rs
+++ b/crates/vize_atelier_vapor/src/generate/operations/refs.rs
@@ -79,32 +79,56 @@ pub(super) fn generate_get_text_child(ctx: &mut GenerateContext, get_text: &GetT
     ctx.push_line_fmt(format_args!("const {} = {}.firstChild", child, parent));
 }
 
-/// Generate ChildRef (_child helper)
+/// Generate ChildRef (`_child` / `_next` / `_nthChild` helper).
+///
+/// Mirrors the Vue Vapor runtime's navigation contract (see
+/// [`build_navigation`]): index 0 is `_child`, index 1 is one `_next` step,
+/// and anything further is an absolute `_nthChild` lookup from the parent.
 pub(super) fn generate_child_ref(ctx: &mut GenerateContext, child_ref: &ChildRefIRNode) {
-    ctx.use_helper("child");
-    if child_ref.offset == 0 {
-        ctx.push_line_fmt(format_args!(
-            "const n{} = _child(n{})",
-            child_ref.child_id, child_ref.parent_id
-        ));
+    let expr = if child_ref.offset > 1 {
+        ctx.use_helper("nthChild");
+        cstr!("_nthChild(n{}, {})", child_ref.parent_id, child_ref.offset)
     } else {
-        ctx.use_helper("next");
-        let expr = build_next_chain(cstr!("_child(n{})", child_ref.parent_id), child_ref.offset);
-        ctx.push_line_fmt(format_args!("const n{} = {}", child_ref.child_id, expr));
-    }
+        ctx.use_helper("child");
+        build_navigation(
+            cstr!("_child(n{})", child_ref.parent_id),
+            child_ref.offset,
+            child_ref.offset,
+            ctx,
+        )
+    };
+    ctx.push_line_fmt(format_args!("const n{} = {}", child_ref.child_id, expr));
 }
 
-/// Generate NextRef (_next helper)
+/// Generate NextRef (`_next` / `_nthChild` helper).
 pub(super) fn generate_next_ref(ctx: &mut GenerateContext, next_ref: &NextRefIRNode) {
-    ctx.use_helper("next");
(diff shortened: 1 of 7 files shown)


ubugeeei:
Note: #3331 by @mymx2 — the reporter of #3330 — fixes the same defect and was opened first. I have disabled auto-merge here so this does not race their contribution.

The two agree on the child-ref half (`_nthChild` for an absolute index). They differ on `generate_next_ref`: #3331 chains bare `_next(expr)` calls, which returns `null` during hydration because the chained calls carry no logical index; this PR routes that case through `_nthChild(parent, index)` as `@vue/compiler-vapor` does, which is correct in both modes. Full analysis in https://github.com/ubugeeei-prod/vize/pull/3331#issuecomment-5113029523.

Maintainer's call which lands. If #3331 is extended with the `nthChild` form for `generate_next_ref`, close this one.

PR #3337, opened 2026-07-29
fix(vapor): give every next() call a real hydration index (#3330)

ubugeeei:
Follow-up to #3331, which fixed the client-side crash @mymx2 reported in #3330. The `_nthChild` route it added for `generate_child_ref` is right and is left untouched here. Two hydration-only defects remain in the `generate_next_ref` half, both invisible to client-side rendering.

The runtime contract:

```js
  return _next(node)
}
```

The second argument is an **absolute index into the parent**, consulted only while hydrating. So:

**1. Chained bare `_next(node)` returns `null` while hydrating.** `build_next_chain` emitted `offset` nested calls with no index for a multi-sibling jump. Each one lands in `locateChildByLogicalIndex(parent, undefined)`, and since `fromIndex === undefined` is never true the loop runs off the end and returns `null`. Now routed through `_nthChild(parent, index)`, the same helper the child-ref case already uses and what `@vue/compiler-vapor`'s `genChildren` does for any distance above one.

**2. A single-step `_next` carried a literal `1`.** That is only correct when the target happens to be the parent's second child. For `<div><p/><p/><span :id="x"/><span :id="y"/></div>`, `y` sits at index 3 one sibling past `x`, and `_next(prev, 1)` resolves index 1 — the wrong node — while hydrating. It now carries the target's real index.

`NextRefIRNode` keeps its public shape: adding the parent and index as fields tripped `cargo-semver-checks`' `constructible_struct_adds_field`, so the generator derives them instead. Every sibling chain is anchored by a `ChildRef` that knows the parent and the absolute index, so each ref operation records where it landed on the crate-private `GenerateContext` and the next hop reads it back.

## Verification

Two new tests fail on current `main` and pass here:

- `single_step_next_carries_the_targets_absolute_index` — pins index `3`, and asserts a literal `1` is not emitted, for the case above.
- `no_bare_next_call_is_emitted` — paren-balanced scan proving every emitted `_next` carries an index (so `_next(_child(n2), 1)` passes and a chained `_next(_next(x))` cannot).

Three tests from the earlier round still pass on both sides, pinning preserved behavior.

On @mymx2's original template the output is unchanged from #3331 — `_nthChild(n1, 3)` then `_next(n2, 1)` — because there the adjacent `<li>` really is at index 1. That is why the reported repro alone does not surface these two.

`cargo test -p vize_atelier_vapor`: 109 passed, 0 failed. fmt and clippy clean.

Part of #3330

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/3337"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved navigation between dynamic sibling elements during hydration.
  - Corrected multi-sibling jumps to use direct child indexing for more reliable runtime behavior.
  - Ensured single-step navigation uses accurate hydration indexes.
  - Prevented invalid chained navigation expressions and step counts.

- **Tests**
  - Added regression coverage for single- and multi-sibling navigation scenarios.


ubugeeei:
Rebased onto current main and fixed both red required checks.

**`test-scripts` (source length)** — `crates/vize_atelier_vapor/src/generate/context.rs` was reported as "over-limit file grew" (441 → 458). That growth came from main moving under the PR, not from this branch; it does not touch that file. The rebase updated the merge-base and the diff no longer contains it.

**`clippy-and-test`** — `stateful_destructured_component_snapshots` in `vize_atelier_jsx`. The only changed byte is the hydration index, and **the old snapshot had pinned the defect**:

```diff
 const t3 = _template("<section><header>…</header><ul></ul><footer>…</footer></section>", true)
 const n0 = _child(n3)          // header, index 0
 const n1 = _next(n0, 1)        // ul,     index 1
-const n2 = _next(n1, 1)        // footer is index 2 — hydration resolves the <ul>
+const n2 = _next(n1, 2)
```

`<footer>` is the section's third child, so the literal `1` would have made `locateChildByLogicalIndex(section, 1)` return the `<ul>` during hydration — the wrong node, silently, while client-side rendering looked fine. That is exactly the second defect this PR is about, so the snapshot needed the update rather than the code.

Refreshed with `INSTA_UPDATE=always` on that single test; nothing else in the snapshot moved.

Green locally after both fixes: `cargo test -p vize_atelier_vapor -p vize_atelier_jsx -p vize_atelier_dom -p vize_atelier_ssr` — 44 suites, 0 failures.

PR #3359, opened 2026-07-30
fix(canon): resolve a symlinked node_modules when composing the virtual root

### LINKED ISSUE #3320: fix(cli): a node_modules symlink redirects project-root discovery and drops real diagnostics
Found while running the class-component oracles from a git worktree (#3298, PR #3318).

## Symptom

When `tests/node_modules` is a **symlink** rather than a real directory, `vize check` resolves a different project root than it should: root discovery walks up to `tests/`, pulls in 83 unrelated files, and a genuinely broken fixture stops being reported. Specifically the injected TS2322 in the `generic-build` fixture disappears.

Removing the symlink and running the *same binary* over the *same fixture* reports the diagnostic correctly. So this is a resolution-context defect, not a type-checking defect — and its failure mode is a **silent false negative**, which is the dangerous direction.

## Why it matters beyond the local setup

Any layout that reaches `node_modules` through a symlink (pnpm-style stores, monorepo hoisting shims, worktree-based CI lanes, containerized checkouts that bind-mount dependencies) can hit the same redirect. The user sees a clean run and has no signal that the project root was wrong.

## Suggested direction

Project-root discovery should either resolve symlinks consistently when walking up, or stop at the first real project marker regardless of link status — and in either case the resolved root and file count are worth surfacing when they differ from the invocation directory's expectation, so a wrong root is visible instead of silent.

## Related

Also found in the same session, test-infra rather than product: `tests/snapshots/check/vue2-elm.ts` cannot pass inside a git worktree at all, because it asserts the pinned fixture SHA before invoking vize and worktrees do not get submodules checked out. Worth handling if worktree-based CI lanes are ever introduced.

ubugeeei:
Closes #3320.

## Change class

Virtual TypeScript and type checking (path resolution).

## Defect

`VirtualProject::new` canonicalized the project root and then joined `node_modules/.vize/canon` onto it:

```rust
let project_root = canonicalize_non_verbatim(project_root);
let virtual_root = project_root.join("node_modules").join(".vize").join("canon");
```

When `node_modules` is itself a symlink, the result traverses that link, so the virtual files are written to the **link target** — physically outside the project root. Corsa reports diagnostics under the target path, mapping them back to the authored SFC fails, and the diagnostics are dropped.

**The failure is silent.** The file is still listed as checked with an empty diagnostics array, so a clean run is indistinguishable from a correct one — the dangerous direction, as #3320 notes.

## Reproduction

Isolated layout: an ancestor `tsconfig.json`, a nested app with no tsconfig of its own, one SFC with a deliberate type error. Same binary, same file, same command; the only variable is whether the ancestor `node_modules` is a real directory or a symlink.

```
node_modules is a real directory:
  "diagnostics": ["error:2:7 [TS2322] Type 'string' is not assignable to type 'number'."]
  "errorCount": 1

node_modules is a symlink:
  "diagnostics": []
  "errorCount": 0
```

Confirmed mechanically: after the symlink run, the virtual project is sitting in the link target — `<store>/.vize/canon/{tsconfig.json, package.json, nested/...}` — not under the project root. The virtual file was generated and checked; only the path mapping was lost.

Layouts that reach `node_modules` through a link are common: pnpm-style stores, monorepo hoisting shims, worktree-based CI lanes, and containers that bind-mount dependencies.

## Fix

Resolve the link when composing the virtual root, so it is already a real path and matches what Corsa reports. `canonicalize_non_verbatim` falls back to the input path on failure, so a project that has no `node_modules` yet keeps today's behavior.

## Tests

Two tests pin **both** layouts, so neither direction can regress silently:

- `virtual_root_resolves_a_symlinked_node_modules_to_its_real_location` (unix-gated) — asserts the virtual root equals `<real store>/.vize/canon`, and asserts it is *not* under the project root, since a symlinked store legitimately sits outside it.
- `virtual_root_stays_inside_the_project_for_a_real_node_modules_directory` — the ordinary layout is unchanged.

Reverting `project.rs` fails the first with the exact path divergence and leaves the second passing:

```
left:  ".../project/node_modules/.vize/canon"
right: ".../store/.vize/canon"
```

## Verification

```sh
cargo test -p vize_canon --lib
cargo fmt --all -- --check
cargo clippy -p vize_canon --all-targets
```

- 618 lib tests pass.
- fmt clean; clippy reports no errors and no new warnings in the touched files.
- The end-to-end repro above now reports the `TS2322` through the symlink, and the real-directory layout still reports it.

## Scope: the LSP is not affected

`overlay_root_for_project` in `crates/vize_canon/src/lsp_client/session_paths.rs` composes its overlay root the same shape:

```rust
project_root.join("node_modules").join(".vize").join("corsa-overlay")
```

so it looked like the same defect. **Tested, and it is not** — driving a real `vize lsp` session over both layouts (`node_modules` as a real directory and as a symlink into an out-of-tree store) publishes `TS2322` in both cases. The LSP maps diagnostics through document URIs rather than a virtual-root path prefix, so the traversed link does not lose them. The batch path is the only affected surface.

That oracle is worth keeping so the LSP cannot regress into this failure mode; it is test-only and lands separately rather than as churn here.

## Not addressed here

#3320 also observes that root discovery walks up past a fixture without its own `tsconfig.json` and adopts an ancestor one (83 unrelated files). That reproduces independently of symlinks and does **not** by itself drop the diagnostic — verified: the injected `TS2322` is still reported when the ancestor `node_modules` is a real directory. Whether root discovery should stop at the first real project marker is a separate behavioral decision; this PR fixes only the silent-false-negative half.

## Follow-up: alias fallbacks outside the project root

Moving the virtual root out of the project tree exposed a containment assumption in `virtual_root_to_project_prefix`: the `strip_prefix` failed, the depth collapsed to zero, and each real-tree alias fallback was emitted as a bare target (`src/*`). TypeScript rejects those with TS5090 ("Non-relative paths are not allowed"), which failed `check_nuxt2_compiler_compat_warning_does_not_suggest_nuxi_prepare` (its fixture symlinks the workspace `node_modules`). The prefix now walks up to the shared ancestor and back down, so the fallback is a real relative route from the store into the project; a contained virtual root still yields the same `../../../`. Pinned by `path_alias_fallbacks_stay_relative_when_node_modules_is_a_symlinked_store`.


ubugeeei:
Converting to draft — CI caught a real problem with this approach, and the fix is wrong as written.

## What broke

`crates/vize/tests/check_nuxt2_fallback_guidance_cli.rs` fails:

```
error:1:1 [TS5090] Non-relative paths are not allowed. Did you forget a leading './'?
```

Reproduced locally, so it is not a CI-environment artifact.

## Why

That fixture builds its project with `link_workspace_node_modules`, which **symlinks the workspace `node_modules` into the fixture root** — exactly the layout this PR targets. With the change, `virtual_root` resolves to `<workspace>/node_modules/.vize/canon`, which is outside the fixture's project root. And `tsconfig_gen.rs` assumes the opposite:

```rust
fn virtual_root_to_project_prefix(&self) -> CompactString {
    let depth = self
        .virtual_root
        .strip_prefix(&self.project_root)
        .map(|relative| relative.components().count())
        .unwrap_or(0);
    // ...
}
```

`strip_prefix` fails, depth falls back to `0`, the alias fallback prefix comes out empty, and the generated tsconfig ends up with non-relative paths — hence TS5090.

## A second problem the test surfaced, worse than the first

If several projects share one store through symlinks — which is the normal case for pnpm and for this test helper — they would now all write their virtual project into the **same** `<store>/.vize/canon`. That is a cross-project collision, not just a broken path prefix. My unit test asserted the virtual root moves outside the project root and called that correct; it is not, and asserting it was me encoding my own assumption rather than checking the design.

## Corrected direction

Do not relocate the virtual root. Keep it nested under the project root (traversing the link, as today) so the tsconfig prefix math and per-project isolation both stay valid, and fix the **reverse mapping** instead: when Corsa reports a diagnostic under the link-target path, resolve it by matching the canonicalized virtual root as well as the literal one.

That keeps writes where they are and only teaches lookup about the second spelling of the same directory — which is where the silent drop actually happens.

The defect in #3320 is real and the reproduction in the PR body stands: with an ancestor `node_modules` symlink, an injected `TS2322` is silently dropped, and the file is still reported as checked with an empty diagnostics array. Only the fix needs redoing.

ubugeeei:
Closing: the approach here is wrong for the reasons in the PR discussion, and the corrected diagnosis is now on #3320 — the diagnostic never reaches the mapping layer at all, so this is a Corsa session/request problem rather than a virtual-root path problem. Reopening a fix under a new branch once that layer is understood.

Issue #3065, opened 2026-07-18
vize canon --declaration drops generic SFC type parameters (TS2304: Cannot find name 'T')


ubugeeei:
## Environment

- vize: 0.291.0 (npm package CLI)
- Corsa runtime: bundled `@typescript/native-preview` (tsgo)
- OS: macOS (darwin arm64), Node 25.8.1

## Summary

`vize check <dir>` passes on a generic SFC, but `vize check <dir> --declaration` fails: the materialized Corsa project seems to lose the `generic` attribute's type parameters, so the generated `*.vue.ts` references `T` outside any scope that declares it.

## Minimal reproduction

`src/Generic.vue`:

```vue
<script setup lang="ts" generic="T extends string = 'a'">
const { value = undefined } = defineProps<{
  value?: T;
}>();

const emit = defineEmits<{
  "update:value": [value: T];
}>();

defineSlots<{
  default: (props: { value: T | undefined }) => unknown;
}>();

function onClick(): void {
  if (value !== undefined) emit("update:value", value);
}
</script>

<template>
  <button type="button" @click="onClick">
    <slot :value="value" />
  </button>
</template>
```

`tsconfig.json`:

```json
{
  "compilerOptions": {
    "target": "ESNext",
    "module": "Preserve",
    "moduleResolution": "Bundler",
    "strict": true,
    "noEmit": true
  },
  "include": ["src"]
}
```

```console
$ vize check src
✓ Type checked 1 files — No type errors found!

$ vize check src --declaration
Building Corsa virtual project for 1 files under …
Error: corsa error (exit code 1): src/Generic.vue.ts(96,29): error TS2304: Cannot find name 'T'.
src/Generic.vue.ts(105,38): error TS2344: Type 'unknown' does not satisfy the constraint 'string'.
```

The same failure occurs with a multi-parameter generic (`generic="Mode extends …, TEvent extends …"`) and with/without default type parameters. Non-generic SFCs in the same project emit `.vue.d.ts` correctly, and the erroring run still emits (broken) declarations for the generic components.

## Expected

`--declaration` emits a `.vue.d.ts` that preserves the component's type parameters, matching what `vize check` (non-declaration) already type-checks successfully.

## Context

Hit in `ubugeeei-prod/vue-calenadar-temporal`, where `CalendarRoot.vue` / `DatePickerRoot.vue` are generic over the selection mode and event type. Currently working around it by overwriting those two emitted files with hand-authored `.vue.d.ts` in the build task.

PR #3365, opened 2026-07-30
fix(canon): instantiate generic module aliases in the generic constructor

### LINKED ISSUE #3354: fix(canon): generic declaration constructors do not instantiate Slots/Emits/Exposed type parameters
### Area

CLI

### Version

0.302.0

### Minimal reproduction

[vize-0.302.0-generic-slots-repro.zip](https://github.com/user-attachments/files/30532181/vize-0.302.0-generic-slots-repro.zip)

I used AI to generate a minimal reproduction example from my project.

### Actual behavior

```ts
export type Slots<T extends string = 'fallback'> = { /* ... */ }

type __VizeGenericComponentConstructor = new <T extends string = 'fallback'>() => {
  $props: __VizeComponentProps<Props<T>>
  $slots: Slots // <---------- there
}
```
It lost generic on slot.

### Expected behavior

```ts
export type Slots<T extends string = 'fallback'> = { /* ... */ }

type __VizeGenericComponentConstructor = new <T extends string = 'fallback'>() => {
  $props: __VizeComponentProps<Props<T>>
  $slots: Slots<T> // <---------- there
}
```

### Environment

- OS: MacOS 15.7.7
- Architecture: Intel i5
- Node.js: v25.9.0
- Package manager: vp@0.2.4 pnpm@11.13.0
- Rust: v1.96.1
- Framework: vite+

### Before submitting

- [x] I checked existing fix reports and pull requests.

ubugeeei:
Closes #3354.

## Change class

Virtual TypeScript and type checking (generic component typing).

## Defect

`__VizeGenericComponentConstructor` declares the SFC's type parameters, so its body has them in scope. `Props` was instantiated there; `Slots`, `Emits`, and `Exposed` were emitted bare:

```ts
type __VizeGenericComponentConstructor = new <T extends string = 'fallback'>(...args: any[]) => {
  $props: __VizeComponentProps<Props<T>> & __EmitProps<Emits>;   // ← Emits bare
  $emit: __EmitFn<Emits>;                                        // ← bare
  $slots: Slots;                                                 // ← bare
} & import('vue').ComponentPublicInstance & __VizeShallowUnwrapRef<Exposed>;  // ← bare
```

Those aliases re-declare the parameters *with defaults*, so a bare reference is legal TypeScript and quietly resolves to the default: a `Slots<T>` keyed on the component's own parameter collapsed to `Slots<'fallback'>`.

**Nothing fails in the virtual module.** There is no diagnostic, no compile error — the wrong types only surface downstream in consumers and in emitted declarations. That is why this needs an output assertion rather than a "does it typecheck" test.

## Fix

Instantiate each alias when it actually took parameters, and leave it bare otherwise. The negative case matters: a generic SFC whose `defineSlots` type never mentions the parameter produces `type Slots = {...}`, and `Slots<T>` against that is a hard `TS2315`. So the decision has to be per-alias, not per-component.

`emit_slots_type` and `emit_exposed_type` now report whether they re-declared the parameters; `EmitsInfo` already tracked this for `__vizeResolveEmitProps`.

The non-generic `__VizeComponentInstance` keeps the bare aliases — no parameters are in scope there, so the declared defaults are correct. Every snapshot below shows those neighbouring lines unchanged.

## Result

```ts
type __VizeGenericComponentConstructor = new <T extends string = 'fallback'>(...args: any[]) => {
  $props: __VizeComponentProps<Props<T>> & __EmitProps<Emits<T>>;
  $emit: __EmitFn<Emits<T>>;
  $slots: Slots<T>;
} & import('vue').ComponentPublicInstance & __VizeShallowUnwrapRef<Exposed<T>>;
```

Multi-parameter components resolve too: `Slots<Mode, TEvent>`, `Exposed<Mode, TEvent>`.

## Tests

Two new tests in the existing `#3065` module, covering both directions:

- `generic_constructor_instantiates_generic_module_aliases` — all four references instantiated, **and** the non-generic instance still bare.
- `generic_constructor_leaves_non_generic_aliases_bare` — a generic SFC whose slots/expose types ignore the parameter must not gain `<T>`, which would be `TS2315`.

Reverting the generator fails the first and leaves the second passing.

## Snapshot refresh

All four diffs are confined to the generic constructor block.

`generic_declaration_emit_outputs` is the strongest evidence because it is the **emitted `.d.ts`**, i.e. what ships to consumers:

```diff
 type __VizeGenericComponentConstructor = new <T extends string = 'a'>(...args: any[]) => {
-    $props: __VizeComponentProps<Props<T>> & __EmitProps<Emits>;
-    $emit: __EmitFn<Emits>;
-    $slots: Slots;
+    $props: __VizeComponentProps<Props<T>> & __EmitProps<Emits<T>>;
+    $emit: __EmitFn<Emits<T>>;
+    $slots: Slots<T>;
 };
```

and for the two-parameter component:

```diff
-    $slots: Slots;
-} & import('vue').ComponentPublicInstance & __VizeShallowUnwrapRef<Exposed>;
+    $slots: Slots<Mode, TEvent>;
+} & import('vue').ComponentPublicInstance & __VizeShallowUnwrapRef<Exposed<Mode, TEvent>>;
```

`virtual_ts_generic_component`, `virtual_ts_generic_slots_component`, and `virtual_ts_multi_generic_exposed_component` show the same lines in the virtual TS.

## Verification

```sh
cargo test -p vize_canon --lib
cargo test -p vize --features legacy --lib
cargo fmt --all -- --check
cargo clippy -p vize_canon --all-targets
```

- 621 canon lib tests pass; 368 CLI lib tests pass.
- fmt clean; clippy 0 errors and no new warnings in the touched files.
- End-to-end: the reported SFC shape now generates all four references instantiated and still typechecks clean.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/3365"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TypeScript typing for generic components.
  * Correctly preserves generic type parameters for component slots, emitted events, exposed values, and constructor instances.
  * Prevents non-generic type aliases from receiving unnecessary type arguments.

* **Tests**
  * Added coverage for generic and non-generic component type aliases to ensure generated typings remain accurate.


ubugeeei:
@codesmith-bot fix CI

PR #3374, opened 2026-07-30
fix(canon): map Corsa CLI diagnostics through a symlinked node_modules

ubugeeei:
Fixes the silent-false-negative half of #3320: `vize check` dropped every diagnostic when the project's `node_modules` is a symlink.

> **Read the "Two tests this un-silences" section before merging.** The fix makes two pre-existing false positives visible again. The one that runs in CI has since been fixed on `main` by #3382, merged in here; the other still needs a decision, not a rubber stamp.

## Reproduction

Ancestor `tsconfig.json`, nested app with no tsconfig of its own, one SFC with a deliberate type error. Same binary, same file, same command; the only variable is whether the ancestor `node_modules` is a real directory or a symlink into an out-of-tree store.

| `node_modules` layout | `diagnostics` | `errorCount` |
| --- | --- | --- |
| real directory | `["error:2:7 [TS2322] Type 'string' is not assignable to type 'number'."]` | 1 |
| symlink to a store | `[]` (before) → same as above (after) | 0 → 1 |

The file was still listed as checked with an empty array, so the run looked clean.

## Root cause (measured, not inferred)

`run_cli_for_config` and `parse_cli_diagnostic_line` were instrumented to print the Corsa CLI's working directory, its raw output, and the virtual path each diagnostic line resolves to. The two layouts diverge in the CLI's own output:

**symlinked**

```
cwd     = <project>/node_modules/.vize/canon          (spelled through the link)
stdout  = ../../../<project-dir>/node_modules/.vize/canon/src/A.vue.ts(30,9): error TS2322: ...
resolved= <project>/<project-dir>/node_modules/.vize/canon/src/A.vue.ts   <- segment doubled
          known=false mapped=false   ->  0 mapped diagnostics
```

**real directory**

```
stdout  = src/A.vue.ts(30,9): error TS2322: ...
resolved= <virtual_root>/src/A.vue.ts
          known=true  mapped=true    ->  1 mapped diagnostic
```

`check_with_cli` starts the CLI with `current_dir(virtual_root)`, and the CLI prints relative diagnostic paths against its working directory. But a process only ever observes the *resolved* spelling of its working directory: with `node_modules` a symlink, `<project>/node_modules/.vize/canon` resolves to `<store>/.vize/canon`, so the CLI computes its relative paths from the link target while `VirtualProject` indexes files under the through-link spelling. `normalize_cli_path` joined the printed path onto the through-link root, the `..` climb out of the link target and back down the project path doubled a segment, `find_by_virtual` missed, `map_to_original` returned `None`, and the diagnostic was discarded.

Nothing surfaced the loss. `run_cli_for_config` deliberately treats a non-zero exit that carries diagnostic-shaped lines as "the CLI ran the project" (falling back to the project-session API there costs orders of magnitude more wall time for the same answer), so a run whose every diagnostic failed to map reports an empty list and looks clean.

## Fix

Resolve a relative CLI diagnostic path against both spellings of the virtual root — the through-link spelling and the resolved spelling the CLI printed against — and keep the candidate that lands inside the virtual project, since that is the only spelling the project's index can be keyed by. A path that resolves to a real file outside the project keeps its resolved spelling; a path that names nothing is left untouched, exactly as before.

The second candidate exists only when the two spellings differ, so a project with a real `node_modules` executes exactly the previous code path.

Extracted to `crates/vize_canon/src/batch/executor/cli/diagnostic_paths.rs` because `cli.rs` is over the 350-line budget (694 → 659).

## Why the two rejected explanations do not work

1. **Canonicalizing the virtual root** (attempted in #3359, closed). `tsconfig_gen.rs::virtual_root_to_project_prefix` derives the alias-fallback prefix from `virtual_root.strip_prefix(project_root)`. Move the root outside the project and `strip_prefix` fails, depth falls back to `0`, the generated tsconfig emits non-relative paths, and `check_nuxt2_fallback_guidance_cli.rs` fails with TS5090. Worse, several projects sharing one store would all write into the same `<store>/.vize/canon`. The virtual root has to stay nested in the project, traversing the link — and it still does here.

2. **Teaching `find_by_virtual` the link-target spelling.** The reported path is neither the through-link nor the link-target spelling of the file: it is a *relative* path that only resolves correctly against the link target, and joining it onto the through-link root produced `.../<project-dir>/<project-dir>/node_modules/...`, which names no file at all. No extra key in `virtual_files` can match that; the path arithmetic has to be corrected before the lookup.

Also worth recording, since the issue thread pointed there: the loss is **not** in the Corsa session/request layer. `map_batch_diagnostics` is never reached on this run — `check_with_servers` only degrades to the project-session API when the CLI *errors*, and here the CLI succeeds. Instrumentation that showed "no iterations at all" was on a path this run never takes.

## Tests

`crates/vize/tests/check_symlinked_node_modules_cli.rs` pins both layouts end to end — identical project, one with `node_modules` symlinked to an out-of-project store, one with a real `node_modules` — and asserts the injected TS2322 is **present** in both, since the dangerous direction is its absence.

`crates/vize_canon/src/batch/executor/cli/diagnostic_paths.rs` adds unit coverage for the path arithmetic: the plain relative case, the symlink case, a file outside the virtual project, and a non-existent path.

Both fail without the fix:

```
test check_reports_diagnostics_with_a_symlinked_node_modules ... FAILED
    symlinked: the injected type error must be reported, not silently dropped
    { "files": [ { "file": "src/Broken.vue", "diagnostics": [] } ], "errorCount": 0, ... }
test check_reports_diagnostics_with_a_real_node_modules ... ok

test ...diagnostic_paths::tests::resolves_relative_paths_reported_against_a_symlinked_virtual_root ... FAILED
  left:  ".../app/node_modules/app/node_modules/.vize/canon/src/App.vue.ts"
  right: ".../app/node_modules/.vize/canon/src/App.vue.ts"
```

`crates/vize/tests/check_nuxt2_fallback_guidance_cli.rs` symlinks the workspace `node_modules` into its fixture, so it exercises this layout; it stays green.

## Two tests this un-silences

Both of these use `link_workspace_node_modules`, which symlinks the whole workspace `node_modules` into the fixture. That is the layout that dropped every diagnostic, so both assertions were vacuous: they have never actually verified anything.

| test | what surfaced | status |
| --- | --- | --- |
| `check_plain_script_named_exports_cli::check_preserves_named_exports_from_normal_script_vue` (**runs in CI**) | `TS2749 'DiffDisplayMode' refers to a value, but is being used as a type here.` | **fixed on `main` by #3382**, merged in |
| `check_nuxt2_plugin_injections_cli::check_nuxt2_false_positive_fixture_matrix` (`--features legacy` only, not in CI) | `TS2339 Property '$emit' does not exist on type '{ clickFirst(): void; inputKey(): void; }'` x3, `TS2339 ... '{ close(): void; }'`, `TS7006`/`TS7031` on `setup(props, { emit })` | still open, needs a decision |

Both are pre-existing product defects, already user-visible for any project with a real `node_modules`. Verified with an **unmodified `origin/main` build**, same fixture content, only the `node_modules` layout differing:

```
named-exports fixture, origin/main binary
  real directory : errorCount 1   ("[TS2749] 'DiffDisplayMode' refers to a value...")
  symlinked      : errorCount 0   <- why the test was green in CI
```

### TS2749, fixed on `main` by #3382

The plain-`<script>` named-export bridge re-exported every binding value-only, so `export const DiffDisplayMode = __vize_plain_script_exports.DiffDisplayMode;` lost the type side of an `enum`.

This branch originally carried its own fix for that: lift a hoistable `export enum` to module scope and drop it from the value bridge. #3382 landed on `main` first with a broader fix, so **that version won the merge and this branch no longer touches the generator at all.** Rather than moving the declaration, #3382 keeps the value bridge and emits a companion type alias next to it:

```ts
export const DiffDisplayMode = __vize_plain_script_exports.DiffDisplayMode;
export type DiffDisplayMode = (typeof DiffDisplayMode)[keyof typeof DiffDisplayMode];
```

That covers strictly more than the hoist did: `const enum`, `class` (via `InstanceType<typeof C>`), and enums with computed members, which the hoist had to leave on the value bridge because a computed member can read a binding that stays inside `__setup()`. It also leaves value-only exports (`const`, `function`) without a type meaning they never had. `main`'s `script_module/plain_exports.rs` and `crates/vize_canon/tests/plain_script_named_exports.rs::normal_script_value_and_type_declarations_are_exported_in_both_spaces` pin all of this.

The merge resolution therefore took `main`'s side wholesale for `script_module.rs` and `plain_script_named_exports.rs` and deleted this branch's `script_module/enum_exports.rs`, so there is only one implementation of the type-side fix in the tree. **The only delta this PR still carries against `main` is the symlinked-`node_modules` diagnostic path mapping**, which `main` never touched:

```
crates/vize/tests/check_symlinked_node_modules_cli.rs | 221 +++
crates/vize_canon/src/batch/executor/cli.rs           |  39 +--
crates/vize_canon/src/batch/executor/cli/diagnostic_paths.rs | 212 +++
```

### Legacy Vue 2 options API, still open

The generated `__VizeThis` bridge lists only `methods`/`props` keys as `any` and omits the instance surface (`$emit`, ...), while the mapped copy of the raw `defineComponent({ ... })` is checked with TypeScript's own `this` for an object literal, i.e. the `methods` object. That defect is a different layer from both this path-resolution fix and the enum bridge, the test does not run in CI, and it should not be papered over by keeping #3320's silent drop. **The call is yours**: land this and track it as its own issue, or fix it first.

## Verification

```
cargo fmt --all -- --check                                        # clean
cargo clippy -p vize_canon --all-targets                          # 0 errors, no warnings in changed/new files
cargo test -p vize_canon --lib                                    # 627 passed; 0 failed
cargo test -p vize --features legacy --test check_symlinked_node_modules_cli
                                                                  # 2 passed
cargo test -p vize --features legacy --test check_nuxt2_fallback_guidance_cli
                                                                  # 1 passed
cargo test -p vize --features legacy --no-fail-fast                # 81 test binaries, 2 targets failed
                                                                  #   -> exactly the two listed above
```

Temporary instrumentation was removed before committing; the diff contains no debug output.

## Residual, out of scope

The other half of #3320 — whether project-root discovery should stop at the first real project marker instead of walking up past a fixture with no tsconfig of its own — is untouched. As recorded in the issue thread, that walk-up reports the injected TS2322 correctly once the path mapping is fixed, so it is an independent decision.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/3374"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved diagnostics for projects using symlinked `node_modules`, including consistent TypeScript error reporting.
  * Fixed diagnostic path handling for virtual projects and files located outside the project.
  * Preserved TypeScript enum type information in generated Vue modules and prevented duplicate runtime exports.

* **Tests**
  * Added regression coverage for symlinked dependencies, diagnostic paths, enum exports, and TypeScript parsing.


ubugeeei:
CI confirms the predicted single failure, and nothing else.

`clippy-and-test` is red only on `check_plain_script_named_exports_cli::check_preserves_named_exports_from_normal_script_vue`:

```
thread 'check_preserves_named_exports_from_normal_script_vue' panicked at
crates/vize/tests/check_plain_script_named_exports_cli.rs:129:5:
  "file": "src/pages/Consumer.vue",
  "diagnostics": [
    "error:6:13 [TS2749] 'DiffDisplayMode' refers to a value, but is being used as a type here. Did you mean 'typeof DiffDisplayMode'?"
  ]
  "errorCount": 1
```

Clippy ran before it and was clean; every other check on the run is green.

That is the pre-existing plain-`<script>` named-export defect described above, not a regression from this change: the same fixture reports the same `TS2749` on an unmodified `origin/main` build as soon as its `node_modules` is a real directory instead of a symlink. Until that defect is fixed — or that assertion is deliberately re-scoped — this PR cannot go green, which is the decision the description asks for.

Issue #3379, opened 2026-07-30
fix(glyph): two idempotence drifts survive the #3346 fix and are different shapes


ubugeeei:
#3368 fixed #3346 and removed all three `idempotence` pins from `tests/_fixtures/glyph-corpus-known-violations.json`. **Two of the three files still drift**, and they are not the same shape as the one that was fixed — so the ledger now claims zero idempotence suppressions while the corpus property will fail.

## Verified on current `main` (`e3000a5c5`)

Two `vize fmt --write --no-config` passes over each pinned file, `--profile ci --features legacy` build:

| fixture | result |
|---|---|
| `dashy` `src/components/Settings/CustomThemeMaker.vue` | **idempotent** — #3368 fixed this one |
| `scalar` `projects/scalar-app/src/features/collection/components/Environment.vue` | **still drifts** |
| `vue-data-ui` `src/components/vue-ui-flow.vue` | **still drifts** |

## They are three different shapes, not one family

#3347 recorded all three as one family and #3368 relied on that. The diffs say otherwise:

**`dashy` (fixed)** — statement-sequence continuation in a directive value. `format_js_expression` cannot parse `foo(1); bar(2)` inside its `void (…)` wrapper, so the `None` arm returned the bytes unanchored. That is what #3368 fixed.

**`scalar` (open)** — interpolation/whitespace boundary, not a directive value at all:
```
<           {{ variable }}
<         </code>
```

**`vue-data-ui` (open)** — a *ternary* continuation, i.e. the path that **parses**:
```
<         ? selectedNodes.includes(path.source) &&
<         selectedNodes.includes(path.target)
```
This is the shape #3368's own regression test asserts is anchored correctly, so something else is re-breaking it here — most likely the `&&` sub-continuation inside the ternary arm.

## Why it went unnoticed

#3368 was developed in a worktree with no hydrated fixtures, so its corpus runs exercised machinery subtests only, not the 33k-file sweep. Only `dashy` was locally reproducible (from #3346's issue text), and the other two were taken on #3347's record. The substitute signal it used — 0/168 non-idempotent across the repo's own `.vue` files — was sound but could not see these two.

Independently corroborated: the #3343 work ran the real corpus over 129 hydrated projects and reported `idempotence 33484 files / 2 waivers used`, i.e. exactly these two files consuming waivers that no longer exist on `main`.

## Immediate action

[...]

Issue #3343, opened 2026-07-29
fix: formatting breaks eslint-disable-next-line when it splits the suppressed line


ubugeeei:
Found while unblocking the release: this is the last `glyph corpus lint-agreement` violation in the CI shard.

## Symptom

`gogocode` `packages/gogocode-vue-playground/packages/vue2/src/components/keycode-modifiers/Comp.vue` reports `a11y/form-control-has-label: 2 -> 3` findings after `vize fmt`.

The pristine source suppresses the third `<input>`:

```vue
<div class="mt20">
  <!--eslint-disable-next-line-->
  custom space: <input v-on:keyup.custom="keys('custom keycode space')" />
</div>
```

The formatter splits that line in two:

```vue
<div class="mt20">
  <!--eslint-disable-next-line-->
  custom space:
  <input @keyup.custom='keys("custom keycode space")' />
</div>
```

"Next line" is now the bare text `custom space:`, so the `<input>` on the line after it is no longer covered and the rule fires.

## Why it matters

Formatting silently disables a user's suppression. The user wrote the disable comment deliberately, ran the formatter, and got a new error with nothing in the diff that looks like it caused one. It also breaks the invariant the lint-agreement property exists to protect: formatting must not change which findings a file produces.

## Two defensible fixes

1. **Formatter**: treat a line preceded by a line-scoped suppression comment as unsplittable, the way whitespace-significant regions are already preserved.
2. **Linter**: anchor `eslint-disable-next-line` to the next *element/statement* rather than the next physical line — the same shape as #3270, which anchored `no-v-html` suppression to the element opening tag.

(2) is more robust because it survives any reflow, but it changes suppression semantics away from ESLint's line-based rule, so it deserves a deliberate decision.

## Reproduction

```sh
cargo build --profile ci -p vize --features legacy
cp tests/_fixtures/_git/gogocode/packages/gogocode-vue-playground/packages/vue2/src/components/keycode-modifiers/Comp.vue /tmp/C.vue
cd /tmp && cp C.vue C.orig && ./target/ci/vize fmt C.vue --write --no-config
diff C.orig C.vue   # the suppressed line is split in two
```

Or run the property: `cd tests && node --test tooling/glyph-corpus-lint-agreement.test.ts`

## Status

Pinned in `tests/_fixtures/glyph-corpus-known-violations.json` so the corpus property stays strict everywhere else while this is decided. Same lifecycle as the #3334 entries: the pin comes out when the defect is fixed.

[...]

Issue #3334, opened 2026-07-29
fix(glyph): formatter re-indents multi-line template literal contents, changing their runtime value


ubugeeei:
**This currently blocks releases.** The v0.301.0 tag run failed `release-preflight / Verify release safety contract` because `Real Project Matrix` failed, and every failing shard failed on the same step: `Check glyph formatter corpus properties` → `glyph corpus parse-preservation holds for every hydrated fixture`.

## Defect

`vize fmt` re-indents the **inside** of a multi-line template literal. Template literal content is significant — the added leading whitespace becomes part of the emitted string at runtime.

From `shadcn-vue` `apps/v4/styles/*/ui/chart/ChartStyle.vue`:

```diff
       ([theme, prefix]) => `
-${prefix} [data-chart=${id}] {
+  ${prefix} [data-chart=${id}] {
 ...
-}
-`,
+  }
+  `,
```

That file builds a CSS string; the reformatted version emits CSS with two extra spaces on every line, and the parse-preservation property sees the `TemplateElement` `cooked`/`raw` values change from `"\n"` to `"\n  "`.

A second family in `vue-termui` (`playground/src/pages/demos/notifications.vue`, `text-selection.vue`) hits the same property through template-position `t\`...\`` attribute values.

## Scope

Locally, with the full fixture corpus hydrated: **97 parse-preservation violations**. CI shard 0 reports 2 (it only hydrates 12 of 130 projects per shard).

## Not a regression from the current release contents

Reproduced with a binary built at `12651dc73` — *before* every PR merged in the current cycle, and immediately after the last glyph fix in this area (#3269 "keep multiline template-literal content verbatim in interpolations"). Byte-identical corruption. So this is a pre-existing defect that the parse-preservation property (#3255) correctly catches; it is not caused by anything in v0.301.0.

#3269 fixed the interpolation path; the script-block path that produces the diff above was not covered by it.

## Reproduction

```sh
cargo build --profile ci -p vize --features legacy
cp tests/_fixtures/_git/shadcn-vue/apps/v4/styles/reka-luma/ui/chart/ChartStyle.vue /tmp/C.vue
cd /tmp && cp C.vue C.orig
./target/ci/vize fmt C.vue --write --no-config
diff C.orig C.vue    # template literal body gains two spaces per line
```

Or run the property directly:

```sh
cd tests && node --test tooling/glyph-corpus-parse-preservation.test.ts
```

## Expected

[...]

PR #3380, opened 2026-07-30
chore(glyph): re-pin the two idempotence drifts that survived #3346

### LINKED ISSUE #3346: fix(glyph): multi-line directive expression continuation lines drift on every glyph pass
The remaining `Real Project Matrix` failure. `glyph corpus idempotence` reports 2 violations, both this shape.

## Symptom

`dashy` `src/components/Settings/CustomThemeMaker.vue` is not idempotent under `vize fmt`. The continuation line of a multi-line directive expression gains two more spaces on **every** pass:

```diff
 @input="customColors[colorName] = $event.target.value;
-              setVariable(colorName, $event.target.value)"
+                setVariable(colorName, $event.target.value)"
```

Source:

```vue
<input
  :value="customColors[colorName]"
  class="swatch-input form__input__element"
  :style="makeSwatchStyles(colorName)"
  @input="customColors[colorName] = $event.target.value;
    setVariable(colorName, $event.target.value)"
/>
```

Unbounded drift, so a project running `vize fmt` in a pre-commit hook re-indents this line further on every commit — the same failure mode as #3334, in the attribute-value path rather than the interpolation path.

## Pre-existing, not from the current cycle

Reproduced with a binary built at `09e2504b4` — the commit **before** both raw-line-mask changes in this cycle (#3338 and #3340) — with byte-identical drift. So neither of those introduced it.

## Reproduction

```sh
cargo build --profile ci -p vize --features legacy
cp tests/_fixtures/_git/dashy/src/components/Settings/CustomThemeMaker.vue /tmp/D.vue
cd /tmp && cp D.vue D1.vue && ./target/ci/vize fmt D1.vue --write --no-config
cp D1.vue D2.vue && ./target/ci/vize fmt D2.vue --write --no-config
diff D1.vue D2.vue    # line 210 gains two spaces
```

Or: `cd tests && node --test tooling/glyph-corpus-idempotence.test.ts`

## Where to look

`compute_raw_line_mask` (`crates/vize_glyph/src/formatter/raw_mask.rs`) deliberately does **not** mark directive-expression continuation lines as raw — its doc comment says they are "formatter output, so they still get SFC indentation unless the value starts on the following line and was preserved verbatim." Here the value starts on the *same* line as the attribute, so the continuation is not treated as verbatim, gets SFC indentation added, and the added spaces are part of the value on the next pass.

Either the continuation line must be re-derived from the expression each pass (so indentation is idempotent), or it must be marked raw like the following-line case already is.

## Status

Pinned in `tests/_fixtures/glyph-corpus-known-violations.json` so the corpus property stays strict elsewhere while this is fixed, same lifecycle as the #3334 and #3343 entries.

ubugeeei:
Refs #3379.

## Why

#3368 fixed #3346 and removed all three `idempotence` pins from `tests/_fixtures/glyph-corpus-known-violations.json`. **Two of the three files still drift.** The ledger therefore claims zero idempotence suppressions while the corpus property will fail on the next Real Project Matrix run.

## Verified on current `main` (`e3000a5c5`)

`cargo build --profile ci -p vize --features legacy`, then two `vize fmt --write --no-config` passes per file:

| fixture | result |
|---|---|
| `dashy` `src/components/Settings/CustomThemeMaker.vue` | **idempotent** — #3368 genuinely fixed this |
| `scalar` `projects/scalar-app/.../Environment.vue` | **still drifts** |
| `vue-data-ui` `src/components/vue-ui-flow.vue` | **still drifts** |

## They are three different shapes

#3347 recorded them as one family and #3368 relied on that. The diffs disagree:

- **`dashy`** — statement-sequence continuation in a directive value; `format_js_expression` cannot parse `foo(1); bar(2)` in its `void (…)` wrapper, so the `None` arm returned bytes unanchored. Fixed.
- **`scalar`** — an interpolation/whitespace boundary, not a directive value at all: the drift is on `{{ variable }}` / `</code>`.
- **`vue-data-ui`** — a **ternary** continuation carrying an `&&` sub-continuation: `? selectedNodes.includes(path.source) &&`. That is the path that *parses*, and the shape #3368's own regression test asserts is anchored correctly — so something else re-breaks it here.

## Why #3368 could not see it

It was developed in a worktree with no hydrated fixtures, so its corpus runs exercised machinery subtests only rather than the 33k-file sweep, and only `dashy` was locally reproducible from #3346's issue text. Its substitute signal (0/168 non-idempotent across the repo's own `.vue` files) was sound but blind to these two.

Corroborated independently: the #3343 work ran the real corpus over 129 hydrated projects and reported `idempotence 33484 files / 2 waivers used` — exactly these two files, consuming waivers that no longer exist on `main`.

## What this PR does

Re-pins both files to #3379, same lifecycle as the #3334 and #3343 entries: the property stays strict everywhere else while the two remaining shapes are fixed on their own merits.

Ledger goes 88 → 90 entries (`parse-preservation` 87, `idempotence` 2, `lint-agreement` 1). **No product change** — fixture data only.

## Note

This is a deliberate re-pin, not a revert of #3368. The `dashy` fix is correct and stays; #3379 tracks the two distinct shapes it did not cover, with explicit warning not to assume they share a cause.

## Summary by CodeRabbit

* **Tests**
  * Added known idempotence issue cases to the test fixture corpus.

### DIFF
--- a/tests/_fixtures/glyph-corpus-known-violations.json
+++ b/tests/_fixtures/glyph-corpus-known-violations.json
@@ -1,5 +1,17 @@
 [
   {
+    "property": "idempotence",
+    "project": "scalar",
+    "path": "projects/scalar-app/src/features/collection/components/Environment.vue",
+    "issue": "https://github.com/ubugeeei-prod/vize/issues/3379"
+  },
+  {
+    "property": "idempotence",
+    "project": "vue-data-ui",
+    "path": "src/components/vue-ui-flow.vue",
+    "issue": "https://github.com/ubugeeei-prod/vize/issues/3379"
+  },
+  {
     "property": "parse-preservation",
     "project": "crater",
     "path": "resources/scripts/components/base/BaseModal.vue",


ubugeeei:
Rebased onto `813a7ba18`. The conflict was #3378 landing its own ledger change: its fix removed the `lint-agreement` pin for #3343, so `main` now carries 87 `parse-preservation` entries only.

Resolved by taking main's ledger and re-adding the two `idempotence` pins on top — **88 → 89** entries (`parse-preservation` 87, `idempotence` 2, `lint-agreement` 0), not 90 as the original PR body said. #3378's removal is correct and preserved; the corpus run behind it reported 0 lint-agreement waivers used across 33486 files.

PR #3381, opened 2026-07-30
fix(canon): mirror per-package node_modules into the virtual project (#3366)

mymx2:
Closes #3366.

## Change class

Virtual TypeScript and type checking (module resolution inside the materialized project).

## Defect

In a pnpm workspace (default isolated linker), a package's own dependencies are linked into `<package>/node_modules`, not the workspace root's. When the effective tsconfig anchors the virtual project at the workspace root, a mirrored file keeps the project-relative shape of its original (`apps/app/vite.config.ts` → `<canon>/apps/app/vite.config.ts`), so Corsa's ancestor walk for a bare specifier never passes the sub-package's real `node_modules`. `<canon>/node_modules` only carries the special-cased `vue`/`vite`/`@vue` set, and the next real ancestor is the workspace-root `node_modules` — the import dies with a spurious TS2307 while `tsgo` on the same tsconfig resolves it.

The issue body has a six-file minimal repro (registry dep only in `apps/app/node_modules` + workspace-root tsconfig → TS2307; same dep at the root → clean).

## Fix

- Compute the real per-package `node_modules` directories from every registered original path (checked sources and passthrough dependencies), skipping the project root itself (kept special-cased) and any ancestor already inside a `node_modules` tree (store-internal paths must not be mirrored back). New `virtual_project/package_node_modules.rs`.
- Symlink each into the mirror at the matching project-relative location and preserve the links across the materialize GC (`preserved_roots`), so stale links from a previous file set are still pruned.
- Share `symlink_path`/`symlink_matches` in `materialize_fs` (moved out of `runtime_deps`).
- Windows: fall back to a **directory junction** when `symlink_dir` is denied — junctions need no privilege on stock Windows (no Developer Mode/elevation), the same reason pnpm links with junctions. Link detection now probes `read_link` instead of `FileType::is_symlink` (which is false for junctions), so junction links round-trip idempotently and stub dirs never write through a link into a real package. Adds the `junction` crate as a Windows-only dependency.

## Tests

- `virtual_project::package_node_modules` unit tests: sub-package mapping, project-root exclusion, `node_modules`-tree exclusion, ancestor dedup.
- `tests/subpackage_registry_dependency.rs` (integration): a workspace-root tsconfig plus a registry dep present only in `apps/app/node_modules` must enforce its contract (`TS2322` on a deliberate misuse) with no `TS2307`, and the mirror link must exist at `<canon>/apps/app/node_modules`.

## Verification

```sh
cargo test -p vize_canon --lib
[...]

PR #3384, opened 2026-07-30
fix(canon): mirror nested package node_modules into the virtual project

### LINKED ISSUE #3366: fix(canon): false TS2307 for registry deps installed only in a sub-package when the tsconfig is at the pnpm workspace root
### Area

CLI / Canon virtual TypeScript

### Version

0.302.0 (also reproduces on 0.291.0)

### Minimal reproduction

A six-file pnpm workspace (default isolated linker):

```text
├─ package.json          # { "name": "repro", "private": true, "type": "module" }
├─ pnpm-workspace.yaml   # packages: [apps/*]
├─ tsconfig.json         # shared tsconfig at the WORKSPACE ROOT
└─ apps/app/
   ├─ package.json       # devDependencies: { "vize": "0.302.0" }  ← only here, NOT at root
   ├─ vize.config.ts     # imports 'vize'
   └─ src/main.ts        # export const answer: number = 42
```

tsconfig.json:

```json
{
  "compilerOptions": {
    "target": "ESNext",
    "module": "ESNext",
    "moduleResolution": "bundler",
    "strict": true,
    "noEmit": true,
    "types": []
  },
  "include": ["apps/**/*.ts"]
}
```

apps/app/vize.config.ts:

```ts
import { defineConfig } from 'vize';

export default defineConfig({
  typeChecker: { enabled: true, tsconfig: '../../tsconfig.json' },
});
```

Steps:

```sh
pnpm install
cd apps/app
./node_modules/.bin/vize check
```

### Actual behavior

```text
Building Corsa virtual project for 2 files under <workspace-root>...
apps/app/vize.config.ts
  error:1:30 [TS2307] Cannot find module 'vize' or its corresponding type declarations.
```

### Expected behavior

No error. The package exists in `apps/app/node_modules`, and `tsgo -p tsconfig.json` on the same tsconfig reports nothing.

### Controls that pass

- `pnpm add -wD vize` (additionally install the same version at the workspace root) → `vize check` passes.
- Point `typeChecker.tsconfig` at a tsconfig inside `apps/app` instead of the workspace root → passes (the canon is then built under `apps/app/node_modules/.vize/canon`, whose ancestor walk finds the package).
- `tsgo`/`tsc` directly on the same tsconfig → passes.

### Root cause (from reading main @ 1ee2237)

1. Canon mirrors checked files under `<project_root>/node_modules/.vize/canon/`, rebasing paths relative to the project root, which follows the tsconfig location. With a workspace-root tsconfig, `apps/app/vize.config.ts` lands at `<canon>/apps/app/vize.config.ts`.
2. Corsa resolves bare specifiers by walking ancestors of the *mirrored* path. `<canon>/node_modules` is populated by `materialize_runtime_dependencies` with only `vue`, `vite`, `@vue`; `prune_runtime_node_modules` deletes everything else (`crates/vize_canon/src/batch/runtime_deps.rs`).
3. The next real ancestor is the workspace-root `node_modules`. A registry dependency linked only in `apps/app/node_modules` (pnpm's isolated default) is never on the mirrored ancestor chain, so it is unresolvable.
4. The gap is masked for `vue`/`vite`/`@vue` (special-cased materialization plus `VIZE_*_PACKAGE` env overrides) and for tsconfig `paths` aliases (carried into the canon tsconfig), so in a real app it typically only fires for config files importing toolchain packages like `vize` itself.

Related: #2954 (closed) covered the hoisted-linker variant of this shape for workspace-linked packages; this report is the isolated-linker + registry-dependency variant and still reproduces on 0.302.0. #3320 (open) is adjacent project-root/symlink territory.

### Suggested direction

Materialize (symlink) each checked package's own `node_modules` into the mirror at the corresponding mirrored directory, or resolve bare specifiers against the importer's original ancestor chain instead of the mirrored one.

### Workaround

Install the affected packages at the workspace root.

### Environment

- Vize: 0.302.0 / 0.291.0 (Corsa runtime via `@typescript/native-preview`)
- OS: Windows 11 x64, Node 24
- pnpm 11.17.0 (`nodeLinker: isolated`)

### Before submitting

- [x] I checked existing fix reports and pull requests (closest: #2954, closed; #3320, adjacent).

ubugeeei:
Fixes #3366.

## Reproduction

A minimal pnpm workspace with the `tsconfig.json` at the **workspace root** and `acme-config` installed **only** under `apps/app`, using pnpm's default isolated linker (a real `apps/app/node_modules` directory holding a symlink into a virtual store):

```text
├─ package.json
├─ pnpm-workspace.yaml
├─ tsconfig.json                    # include: ["apps/**/*.ts"]
├─ node_modules/                    # empty; the canon is written here
└─ apps/app/
   ├─ package.json                  # devDependencies: { "acme-config": "1.0.0" }
   ├─ vize.config.ts                # import { defineConfig } from "acme-config"
   ├─ src/main.ts
   └─ node_modules/
      ├─ .pnpm/acme-config@1.0.0/node_modules/acme-config/   # real package
      └─ acme-config -> .pnpm/acme-config@1.0.0/node_modules/acme-config
```

Reference tools on the identical workspace, same tsconfig:

```console
$ tsgo -p tsconfig.json      ; echo $?
0
$ vue-tsc -p tsconfig.json --noEmit ; echo $?
0
```

`vize check` before this change:

```console
$ cd apps/app && vize check --tsconfig ../../tsconfig.json
Building Corsa virtual project for 2 files under <workspace-root>...
<workspace-root>/apps/app/vize.config.ts
  error:1:30 [TS2307] Cannot find module 'acme-config' or its corresponding type declarations.
✗ 1 error(s)
```

After this change: `✓ No type errors found!` — while an import of a package that is genuinely not installed still reports `TS2307` at the right position.

## Root cause

`VirtualProject` mirrors checked files under `<project_root>/node_modules/.vize/canon`, rebasing each path relative to the project root — which follows the tsconfig location. With a workspace-root tsconfig, `apps/app/vize.config.ts` lands at `<canon>/apps/app/vize.config.ts`.

Corsa then resolves bare specifiers by walking the `node_modules` directories on the ancestor chain of that **mirrored** path:

| ancestor | `node_modules` present? |
| --- | --- |
| `<canon>/apps/app` | no |
| `<canon>/apps` | no |
| `<canon>` | yes — but `runtime_deps::prune_runtime_node_modules` keeps only `vue`, `vite`, `@vue` |
| `<project_root>` | yes — Node's algorithm skips `node_modules` path components, so the walk jumps here directly |

Every real `node_modules` *between* the project root and the importing file — exactly where pnpm's isolated linker puts a sub-package's own dependencies — is absent from the mirrored chain. The package is therefore unresolvable, and `TS2307` fires even though the dependency is installed and both reference checkers resolve it.

This is invisible in a single-package project (the project root's own `node_modules` is on the chain), and masked for `vue`/`vite`/`@vue` (special-cased materialization) and for tsconfig `paths` aliases (re-anchored into the canon tsconfig), which is why it typically only surfaces for config files importing toolchain packages.

## The fix

`crates/vize_canon/src/batch/virtual_project/package_node_modules.rs` computes, from the materialized file set, every ancestor directory below the project root that has a real `node_modules`, and `materialize()` links each one into the mirror at its rebased path. The mirrored ancestor chain then matches the real one, so resolution is TypeScript's ordinary Node resolution — no `TS2307` suppression anywhere, and a genuinely missing module still errors.

Two safety properties, both covered by unit tests:

- **The prune never descends into a real install.** The mirrored `node_modules` paths join `<canon>/node_modules` as preserved roots for `prune_unexpected_entries`, so the cache GC cannot walk through a link into the user's dependencies. (`std::fs::remove_dir_all` does not follow symlinks, so a stale mirrored directory is unlinked rather than recursed into.)
- **No write inside the virtual project can land in a real install.** If the mirror materializes files *under* a nested `node_modules` (possible when the file walker descends into a real, non-gitignored, non-symlinked dependency directory), that path keeps its mirrored copies and is not linked. Behavior there is unchanged from today; the link never shadows a path the mirror writes to.

The ambient stub writers moved from `materialize.rs` to a new `materialize_stubs.rs` — `materialize.rs` was at 347 lines and could not grow past the 350-line budget. No behavior change in that move.

## Tests

`tests/tooling/cli-check-sub-package-dependency.test.ts` builds the pnpm-shaped workspace above and runs the real CLI:

1. **Primary** — asserts `TS2307` is *absent* on `vize.config.ts` and `errorCount === 0`. The failure mode is a spurious diagnostic, so this asserts absence.
2. **Control** — same layout, plus an import of `totally-not-installed`. Asserts *exactly one* `TS2307`, that it names `totally-not-installed`, and that no diagnostic mentions `acme-config`. This is what stops the test from passing if resolution errors were suppressed wholesale.

The root `node_modules` is a real directory and the sub-package `node_modules` is a real directory containing a per-package symlink, so nothing puts the generated canon behind a link.

`package_node_modules.rs` adds four unit tests: links a sub-package `node_modules`; skips the project root and directories without one; keeps mirrored dependency copies instead of linking over them; links every ancestor that carries its own `node_modules`.

### Verified the tests fail without the fix

Reverted only the `materialize_package_node_modules(&package_links)` call (leaving the module and the test in place), rebuilt, reran:

```text
✖ a dependency installed only in a sub-package resolves under a workspace-root tsconfig
  actual: ["error:1:30 [TS2307] Cannot find module 'acme-config' ..."]  expected: []
✖ a module that is genuinely not installed still reports TS2307 in the same layout
  2 !== 1   (both 'acme-config' and 'totally-not-installed' reported)
```

Both pass again with the call restored.

## Verification

| command | result |
| --- | --- |
| `cargo test -p vize_canon --lib` | **636 passed, 0 failed** |
| `cargo test -p vize --features legacy --no-fail-fast` | **591 passed, 1 failed** — `check_nuxt2_false_positive_fixture_matrix`, **pre-existing**: reproduced identically with the source change reverted and rebuilt. **0 new failures.** |
| `cargo fmt --all -- --check` | clean |
| `cargo clippy --workspace --all-targets` | no warnings from any file this PR touches or adds |
| compat ratchet (below) | **12/12 pass**, every probe holds |

Compat ratchet, all nine probe fixtures hydrated:

```console
$ VIZE_TEST_VUE_TSC=node_modules/.pnpm/vue-tsc@3.3.4_typescript@6.0.3/node_modules/vue-tsc/bin/vue-tsc.js \
  VIZE_TEST_BIN=target/ci/vize node --test tests/tooling/compat-ratchet.test.ts
create-vue:        shared=0  documented=0 falsePositives=0 falseNegatives=0
element-plus:      shared=3  documented=0 falsePositives=0 falseNegatives=0
misskey:           shared=8  documented=0 falsePositives=0 falseNegatives=0
nuxt-ui:           shared=11 documented=1 falsePositives=0 falseNegatives=0
pinia:             shared=7  documented=0 falsePositives=0 falseNegatives=0
vue-router:        shared=0  documented=0 falsePositives=0 falseNegatives=0
vue-element-admin: shared=0  documented=0 falsePositives=0 falseNegatives=0
vitepress:         shared=2  documented=0 falsePositives=0 falseNegatives=0
vue-vben-admin:    shared=16 documented=0 falsePositives=0 falseNegatives=0
ℹ pass 12   ℹ fail 0
```

**`compat-baseline.json` is deliberately not regenerated.** The run reports that `vue-element-admin` improved (`falsePositiveCount` 1 → 0), but that improvement is **not** caused by this PR: it reproduces byte-for-byte with the source change reverted and the binary rebuilt, so the baseline was already stale on `main`. Tightening it here would attribute an unrelated improvement to this change. It should be tightened in whichever PR actually earned it.

## Explicitly not confirmed

- Verified on macOS only. The mirroring reuses `runtime_deps`'s existing `symlink_path`, which already handles the Windows `symlink_dir`/`symlink_file` split, and linking is best-effort: a platform that refuses the symlink degrades to today's resolution behavior rather than failing the check. Windows CI is the first real exercise of that path.
- The LSP uses a different overlay root (`node_modules/.vize/corsa-overlay`) and is untouched by this change; whether the same gap exists there is #3320 territory and out of scope here.
- One corner is knowingly left as-is: if a nested `node_modules` is a *real* directory whose files the input walker sweeps in as check inputs (needs no `.gitignore` covering `node_modules` and no package symlinks — i.e. neither pnpm's nor npm's normal output), the mirrored copies win and the link is skipped, so `TS2307` can still fire there. Linking over those paths would make virtual-project writes land in the user's real `node_modules`, which is not an acceptable trade. Worth noting for a follow-up: `tsc` does not treat such files as root inputs at all (verified with `tsgo --listFiles`), so vize collecting them is itself a divergence — a separate defect from this one.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/3384"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dependency resolution for nested packages and pnpm-style workspaces.
  * Prevented false “Cannot find module” diagnostics for dependencies installed in sub-package `node_modules` directories.
  * Preserved correct errors for genuinely missing packages.
  * Improved generation and updating of TypeScript declaration stubs, including framework and Vue-related helpers.

### DIFF
--- a/crates/vize_canon/src/batch/runtime_deps.rs
+++ b/crates/vize_canon/src/batch/runtime_deps.rs
@@ -147,6 +147,14 @@ fn ensure_stub_dir(path: &Path) -> std::io::Result<()> {
     Ok(())
 }
 
+/// Link `target` to the canonicalized `source` package directory.
+///
+/// Shared with the nested-package `node_modules` mirror so both mirrors resolve
+/// through the same canonicalization and idempotence rules.
+pub(super) fn symlink_package_dir(source: &Path, target: &Path) -> std::io::Result<()> {
+    symlink_path(&package_link_source(source), target)
+}
+
 fn symlink_path(source: &Path, target: &Path) -> std::io::Result<()> {
     if symlink_matches(source, target)? {
         return Ok(());
--- a/crates/vize_canon/src/batch/virtual_project.rs
+++ b/crates/vize_canon/src/batch/virtual_project.rs
@@ -43,6 +43,8 @@ mod jsx_build;
 mod jsx_codegen;
 mod mapping;
 mod materialize;
+mod materialize_stubs;
+mod package_node_modules;
 mod passthrough;
 mod paths;
 mod project;
--- a/crates/vize_canon/src/batch/virtual_project/materialize.rs
+++ b/crates/vize_canon/src/batch/virtual_project/materialize.rs
@@ -5,8 +5,7 @@
 use std::path::{Path, PathBuf};
 
 use rayon::prelude::*;
-use serde_json::Value;
-use vize_carton::{FxHashSet, String as CompactString, profile};
+use vize_carton::{FxHashSet, profile};
 
 use crate::batch::error::CorsaResult;
 use crate::batch::materialize_fs::{
@@ -14,9 +13,12 @@ use crate::batch::materialize_fs::{
 };
 use crate::batch::runtime_deps::materialize_runtime_dependencies;
 
+use super::package_node_modules::{
+    PackageNodeModulesLink, materialize_package_node_modules, package_node_modules_links,
+};
 use super::{
-    AUTO_IMPORT_STUBS_FILE, MODULE_AUGMENTATION_STUB_PREFIX, MODULE_AUGMENTATION_STUBS_FILE,
-    PACKAGE_BOUNDARY_FILE, SHARED_HELPERS_FILE, VUE_MODULE_STUBS_FILE, VirtualProject,
+    AUTO_IMPORT_STUBS_FILE, MODULE_AUGMENTATION_STUBS_FILE, PACKAGE_BOUNDARY_FILE,
+    SHARED_HELPERS_FILE, VUE_MODULE_STUBS_FILE, VirtualProject,
 };
 
 impl VirtualProject {
@@ -31,6 +33,7 @@ impl VirtualProject {
     /// so TypeScript's own filesystem caches are not invalidated.
     pub fn materialize(&self) -> CorsaResult<()> {
         let expected_files = self.expected_materialized_files();
+        let package_links = self.package_node_modules_links(&expected_files);
         profile!(
             "canon.project.prepare_dir",
             ensure_materialize_root(&self.virtual_root)
@@ -41,7 +44,7 @@ impl VirtualProject {
             prune_unexpected_entries(
                 &self.virtual_root,
                 &expected_files,
-                &[self.virtual_root.join("node_modules")]
+                &self.preserved_prune_roots(&package_links)
             )
         )?;
 
@@ -49,6 +52,11 @@ impl VirtualProject {
             "canon.project.runtime_deps",
             materialize_runtime_dependencies(&self.project_root, &self.virtual_root)
         )?;
+
+        profile!(
+            "canon.project.package_deps",
+            materialize_package_node_modules(&package_links)
+        );
 
         profile!(
             "canon.project.write_package_boundary",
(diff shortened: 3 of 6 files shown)


mymx2:
Closed my overlapping #3381 in favor of this — the mirrored-copies-win carve-out and the anti-suppression control are both worth having.

One gap to flag for stock Windows (no Developer Mode, unelevated shell): `symlink_dir` is denied there, so the new mirror links silently degrade to pre-fix resolution — exactly the shape in the "Explicitly not confirmed" section. #3381 had a working version of that piece:

- `junction` crate fallback in the shared `symlink_path` when `symlink_dir` fails — directory junctions need no privilege, same reason pnpm links with junctions on Windows
- `read_link` probing instead of `FileType::is_symlink` (false for junctions) so links round-trip idempotently instead of being torn down and recreated every run
- a matching probe in the stub-dir path so ambient stubs are never written through a link into a real package
- side effect: the existing `vue`/`vite`/`@vue` runtime-deps links also materialize on Windows instead of degrading to stubs

Verified locally on Windows 11 against the #3366 repro (mirror link materializes as a junction and the dependency's contract is enforced).

Happy to rebase just that part onto this branch as a small follow-up PR once this lands.

Issue #3363, opened 2026-07-30
perf(vite): pre-compilation is ~37% of build time and repeats in full on every cold build


ubugeeei:
Measured baseline from the existing harness so the Vite/Nuxt build-performance work starts from numbers rather than impressions.

## Baseline

`node bench/vite.ts 300` (300 SFCs, 1.2 MB, 15 cores):

```
@vitejs/plugin-vue  :  298ms  (1.0k files/s)
@vizejs/vite-plugin :  212ms  (1.4k files/s)  1.4x faster
```

Pre-compilation inside the measured run:

```
[vize] Pre-compiling 300 Vue files... (300 changed, 0 cached, 0 removed)
[vize] Pre-compilation complete: 300 recompiled, 0 reused, 0 failed (78.38ms, native batches: 62.53ms)
```

So on this corpus:

- pre-compilation is **78ms of a 212ms build — about 37%**;
- of that, **62.5ms is the native batch** and roughly **16ms is JavaScript-side overhead** around it.

## Opportunity 1: the pre-compile cache is process-local only

`state.precompileMetadata` is an in-memory `Map` keyed by absolute path with `{ mtimeMs, size }`. That is correct and it works within a session — a warm rebuild logs `cache reused` and skips the batch entirely.

But it is empty on every process start, so every cold build recompiles everything. That is the common case in CI, in `vite build`, and on each dev-server start. On this corpus that is 78ms; it scales with file count.

A persistent on-disk cache would remove that, and the metadata shape is already the right one. The design question is invalidation — it has to key on at least the plugin version, the resolved compiler options, and the Vue dialect, or a stale entry silently serves wrong output. Worth doing deliberately rather than quickly, since the failure mode is incorrect compilation rather than slowness.

Note for anyone benchmarking this: `bench/vite.ts` runs a warmup build and then the measured build in separate Vite instances, so the measured run legitimately reports `0 cached`. That is the harness, not a cache bug.

## Opportunity 2: the metadata scan is a serial sync syscall loop

`precompile` globs the scan patterns and then walks the result calling `fs.statSync` on the main thread, one file at a time, before any compilation starts:

```ts
for (const file of sfcFiles) {
  const stat = fs.statSync(file);
  currentMetadata.set(file, { mtimeMs: stat.mtimeMs, size: stat.size });
}
```

[...]

PR #3409, opened 2026-07-30
fix(bench): stop measuring the Nuxt build against shared warm caches

ubugeeei:
## The defect

`prepareNuxtDir` gives every generated Nuxt app a fresh directory under `target/tool-benchmark/nuxt/`, but it symlinks each app's `node_modules` at **one shared directory** so the app can resolve `nuxt` and `vue` (`bench/compare-tools.mjs:461-466`):

```js
const nuxtNodeModules = join(rootDir, "npm", "framework/nuxt", "node_modules");
symlinkSync(existsSync(nuxtNodeModules) ? nuxtNodeModules : join(benchDir, "node_modules"),
            join(outputDir, "node_modules"), "dir");
```

Every build cache that lives under `node_modules` is therefore shared by every measured run, by both variants, and **across separate invocations of the benchmark** — `target/tool-benchmark` is wiped at startup, that directory is not.

The one that breaks the comparison is `node_modules/.vize/vite-precompile/`, `@vizejs/vite-plugin`'s persistent pre-compile manifest (#3372, #3387). Only the `@vizejs/nuxt` variant can use it, so every measured run after the first restores its modules from disk while the Nuxt default compiler recompiles.

This is the same class of defect as #3392, in a different harness.

## Proof

Machine: Apple Silicon, 12 cores, macOS 26.5.1, node v24.14.0.

```sh
rm -rf npm/framework/nuxt/node_modules/.vize
nix develop -c node bench/compare-tools.mjs --input bench/__in__ --tasks nuxt \
  --nuxt-file-count 20 --runs 2 --warmups 1
stat -f "%m %N" npm/framework/nuxt/node_modules/.vize/vite-precompile/*.vpc
# 1785423491 npm/framework/nuxt/node_modules/.vize/vite-precompile/41ade7e49080080dc74ebc57a25cf06b.vpc

nix develop -c node bench/compare-tools.mjs --input bench/__in__ --tasks nuxt \
  --nuxt-file-count 20 --runs 1 --warmups 0
stat -f "%m %N" npm/framework/nuxt/node_modules/.vize/vite-precompile/*.vpc
# 1785423491 npm/framework/nuxt/node_modules/.vize/vite-precompile/41ade7e49080080dc74ebc57a25cf06b.vpc
```

The mtime is unchanged after a second, fully independent invocation. The cache is only rewritten when something was compiled and staged, so an unchanged mtime means **nothing was recompiled** — the second invocation's Vize build was served entirely by the manifest the first one wrote.

Two more caches share that directory and are worth clearing for the same reason, though they are symmetric between the variants: `.cache/` (Nuxt's and jiti's, observed as `node_modules/.cache/{jiti,nuxt}` after a run) and `.vite/` (Vite's dependency optimizer). Whichever variant runs first pays to populate them.

## The change

Every measured Nuxt build clears `.vize`, `.cache`, and `.vite` under the resolved `node_modules`, then **asserts** they are gone, so the harness cannot drift back. Verified after the change: the manifest is recreated on the last build of the run, which only happens if it was removed before each one.

```
$ nix develop -c node bench/compare-tools.mjs --input bench/__in__ --tasks nuxt \
    --nuxt-file-count 20 --runs 3 --warmups 1
| Nuxt default compiler | 5.43s | 4 files/s | 4.12s, 5.76s, 5.43s |
| @vizejs/nuxt          | 5.10s | 4 files/s | 5.10s, 6.26s, 4.14s |
```

Those two lanes are within noise of each other at 20 files on a machine under heavy competing load — 20 SFCs is a rounding error inside a ~5 s Nuxt build, so this run is evidence that the harness works, **not** a replacement number for the published one.

## What this means for the published Nuxt row

`README.md:69` and `docs/content/architecture/performance-blacksmith.md:28` publish the Nuxt row as `0.9x` / `1.1x` respectively. Whatever the correct value, it was measured with `@vizejs/nuxt` warm on every run but the first. **The correction goes against Vize**: the honest cold-vs-cold figure is worse than what is published, not better.

Regenerating the snapshot on the Blacksmith runner is a follow-up to this PR, not part of it — the published numbers must not be refreshed from a harness that is still on the old behaviour.

## Not affected

The Vite lane in the same file: `prepareViteDir` creates a work directory with **no** `node_modules`, so the plugin's manifest lands inside `target/tool-benchmark/vite/<label>-NNNN/node_modules/.vize` and is wiped with `workRoot` at startup. Every Vite run there already starts cold.

## Tests

`tests/tooling/benchmark-nuxt-cache-state.test.ts`, six cases over the extracted `bench/nuxt-build-cache.mjs`, with `assert.deepEqual` over full lists and `assert.equal` on the complete refusal message. The fixture reproduces the benchmark's own layout: several app directories whose `node_modules` are symlinks to one shared tree.

- the shared-cache list is exactly `[".vize", ".cache", ".vite"]`;
- `node_modules` is reported at its real path, so the sharing is visible rather than hidden behind the link;
- an app without `node_modules` has nothing to read, clear, or assert;
- **the defect itself**: a cache written through one app is visible through every other app;
- clearing through one app clears it for all of them and leaves the installed packages untouched;
- asserting cold against a warm shared cache throws, naming what is left.

Temp trees are created under `os.tmpdir()` and no build artifact is asserted to exist, so this runs from a clean checkout.

```
$ nix develop -c node --test tests/tooling/benchmark-nuxt-cache-state.test.ts
ℹ tests 6  ℹ pass 6  ℹ fail 0
$ nix develop -c node --test tests/tooling/tool-benchmark.test.ts
ℹ tests 4  ℹ pass 4  ℹ fail 0
```

## Gates

```
$ nix develop -c vp run fmt:check      # exit 0
$ nix develop -c vp run source:lengths # exit 0
```
No Rust changed. Based on `origin/main` after #3394, and touches only the Nuxt measurement region of `compare-tools.mjs` (not the reporting split).

Refs #3363

### DIFF
--- a/bench/compare-tools.mjs
+++ b/bench/compare-tools.mjs
@@ -15,7 +15,6 @@ import {
   readFileSync,
   rmSync,
   statSync,
-  symlinkSync,
   writeFileSync,
 } from "node:fs";
 import { basename, delimiter, dirname, join, parse, relative, resolve, sep } from "node:path";
@@ -31,6 +30,7 @@ import {
   renderSurfaceTable,
 } from "./compare-tools-report.mjs";
 import { createNativeBatchSequenceVariants, measureNativeBatchCompile } from "./native-batch.mjs";
+import { linkColdNodeModules } from "./nuxt-build-cache.mjs";
 
 export { createSurface } from "./compare-tools-report.mjs";
 
@@ -467,11 +467,10 @@ function prepareNuxtDir(inputDir, files, label, invocation, useVize) {
   for (const file of files) {
     copyFileSync(join(inputDir, file), join(outputDir, "components", file));
   }
-  const nuxtNodeModules = join(rootDir, "npm", "framework/nuxt", "node_modules");
-  symlinkSync(
-    existsSync(nuxtNodeModules) ? nuxtNodeModules : join(benchDir, "node_modules"),
-    join(outputDir, "node_modules"),
-    "dir",
+  linkColdNodeModules(
+    outputDir,
+    [join(rootDir, "npm", "framework/nuxt", "node_modules"), join(benchDir, "node_modules")],
+    label,
   );
 
   const imports = [];
--- a/bench/nuxt-build-cache.mjs
+++ b/bench/nuxt-build-cache.mjs
@@ -1,0 +1,109 @@
+/**
+ * Cache-state control for the Nuxt benchmark.
+ *
+ * `prepareNuxtDir` gives every generated app a fresh directory under
+ * `target/tool-benchmark/nuxt/`, but it symlinks each app's `node_modules` at
+ * one shared directory (`npm/framework/nuxt/node_modules`, falling back to
+ * `bench/node_modules`) so the app can resolve `nuxt` and `vue`. Every build
+ * cache that lives under `node_modules` is therefore shared by every measured
+ * run, by both variants, and across separate invocations of the benchmark --
+ * `target/tool-benchmark` is wiped at startup, but this directory is not.
+ *
+ * Two things live there and both distort the measurement:
+ *
+ * - `.vize/vite-precompile/` is `@vizejs/vite-plugin`'s persistent pre-compile
+ *   manifest. Only the `@vizejs/nuxt` variant can use it, so every measured run
+ *   after the first restores its modules from disk while the Nuxt default
+ *   compiler recompiles. Measured: after one benchmark invocation, a second
+ *   invocation with `--warmups 0 --runs 1` left the manifest's mtime unchanged,
+ *   which only happens when nothing was recompiled.
+ * - `.cache/` (Nuxt's and jiti's build caches) and `.vite/` (Vite's dependency
+ *   optimizer) are shared by *both* variants, so whichever runs first pays to
+ *   populate them and the other does not.
+ *
+ * Clearing all three before each measured build puts both variants in the same
+ * state on every run, which is the only state in which their times can be
+ * compared.
+ */
+
+import { existsSync, readdirSync, realpathSync, rmSync, symlinkSync } from "node:fs";
+import { join } from "node:path";
+
+/**
+ * Build caches under a Nuxt app's `node_modules`, relative to it.
+ *
+ * `.vize` is the pre-compile manifest one variant owns; `.cache` and `.vite`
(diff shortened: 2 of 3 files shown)


ubugeeei:
Updated after CI went red on the source-length gate.

`bench/compare-tools.mjs` is already over the 350-line budget, so the allowlist permits it to exist but not to grow, and the inline clearing added 8 lines. Rather than grandfather that growth, the clearing moved into the module that already existed for it:

- `linkColdNodeModules(appDir, candidates, label)` in `bench/nuxt-build-cache.mjs` now creates the symlink, clears `.vize`/`.cache`/`.vite`, and asserts they are gone. The symlink is what makes those caches shared in the first place, so a caller that creates one can no longer forget the clear.
- `prepareNuxtDir`'s six-line `symlinkSync` block became a single call, and `symlinkSync` is no longer imported there.

`bench/compare-tools.mjs` is now at **1381 lines, exactly its `origin/main` base** — no growth. Rebased on current `origin/main` (past #3412 and #3415).

```
$ nix develop -c node --test tests/tooling/source-file-lengths.test.ts
✔ source length script checks the current checkout
ℹ pass 7  ℹ fail 0
$ nix develop -c node --test tests/tooling/benchmark-nuxt-cache-state.test.ts
ℹ pass 7  ℹ fail 0     (a seventh case now covers linkColdNodeModules)
$ nix develop -c vp run fmt:check    # exit 0
```

Re-verified the harness still runs after the restructure:

```
$ rm -rf npm/framework/nuxt/node_modules/.vize npm/framework/nuxt/node_modules/.cache
$ nix develop -c node bench/compare-tools.mjs --input bench/__in__ --tasks nuxt \
    --nuxt-file-count 20 --runs 1 --warmups 1
| Nuxt default compiler | 4.99s | 4 files/s | 4.99s |
| @vizejs/nuxt          | 6.70s | 3 files/s | 6.70s |
```

Single run, 20 files, machine under heavy competing load — evidence the harness works, not a number to quote.

Provenance for every measurement in this PR: `npm/native/vize-vitrine.darwin-arm64.node`, release build, sha256 `c7b518d75f1ddd2c8e06c00331845bf3881805137f542fb2f13e1eb95f5e3be4`, built in this worktree and unchanged throughout.

Issue #3391, opened 2026-07-30
feat(atelier): opt-in @vue/babel-plugin-jsx full-compatibility mode for JSX/TSX


ubugeeei:
## Goal

Add an **opt-in full-compatibility mode with `@vue/babel-plugin-jsx`** for Vize's JSX/TSX
compilation, so a project that builds today with `@vue/babel-plugin-jsx` gets the same
*semantics* from Vize. Default Vize behavior must not change — that would be a silent
compatibility break for existing Vize users.

`crates/vize_atelier_jsx/tests/PARITY_INVENTORY.md` states the current position explicitly:
the existing parity tests assert *Vize's* output structure, not babel parity, because Vize
emits through `vize_atelier_dom` / `vize_atelier_vapor`. This issue is about closing that gap
behind a switch.

## Differential oracle

A real differential oracle now exists (see the first PR in the series):

- `crates/vize_atelier_jsx/tests/babel_compat/fixtures/corpus.json` — 94 cases across
  options / elements / props / events / directives / slots / children / optimize / errors.
- `crates/vize_atelier_jsx/tests/babel_compat/oracle.mjs` — runs each case through the real
  `@vue/babel-plugin-jsx` (pinned 2.0.1, catalog `babel-jsx-oracle`) and records the output.
- `crates/vize_atelier_jsx/tests/babel_compat/fixtures/babel-output.json` — the committed
  ground truth, re-derived from the installed plugin in CI by
  `tests/tooling/babel-jsx-oracle.test.ts`.
- `crates/vize_atelier_jsx/tests/babel_compat_oracle.rs` — snapshots babel's output beside
  Vize's per category with an explicit per-row verdict.
- `crates/vize_atelier_jsx/tests/BABEL_COMPAT_INVENTORY.md` — the written inventory.

Result: **63 of 94 rows are already semantically equivalent, 29 diverge, 2 are deferred.**

## Global divergences (apply to nearly every row)

1. **Module shape.** Babel rewrites the JSX expression in place, leaving `const A = () => …`
   intact. Vize emits a standalone `export function render(_ctx, _cache)` for arrow-body
   components. A drop-in compat mode has to preserve the surrounding module.
2. **Block tree.** Babel emits bare `createVNode` and never opens a block. Vize emits
   `openBlock()` + `createElementBlock` / `createBlock`.
3. **Patch flags are inverted.** Babel's **default** is `optimize: false` — *no* patch flags
[...]

PR #3441, opened 2026-07-30
feat(atelier)!: add an opt-in @vue/babel-plugin-jsx compatibility mode

ubugeeei:
Refs #3391. **Draft** — the switch and its plumbing, deliberately with no behavior
attached yet.

## What this adds

The opt-in `@vue/babel-plugin-jsx` compatibility switch, off by default, plumbed
through the same path `jsxMode` already takes:

| Layer | Surface |
| --- | --- |
| compiler | `vize_atelier_jsx::JsxCompatMode` (`Native` \| `Babel`), `JsxCompileConfig::compat` |
| config | `compiler.jsxCompat: "native" \| "babel"`, `ConfigFeatureFlags::jsx_compat`, `load_compiler_jsx_compat()` |
| native | NAPI `compileJsx({ jsxCompat })`, wasm `jsxCompat` option |

Still to plumb before this leaves draft: the JSON schema + generated types
(`vp run gen:schema` / `gen:types`), and the Vite / unplugin / rspack option
surfaces. Grep `jsxMode` for the remaining sibling entries.

## Design decisions, written down so they are not relitigated

**Off by default, proven by a test.** `JsxCompatMode::default()` is `Native`, and
`default_config_output_equals_explicit_native` asserts that `Default::default()` and an
explicit `Native` emit the same module byte for byte across a source touching elements,
props, interpolation, control flow, directives and slots. Turning compat on by default
would be a silent compatibility break for every existing Vize user.

**No per-component directive prologue.** `JsxOutputMode` can be selected per component
with `"use vue:vapor"` / `"use vue:vdom"` because VDOM and Vapor components coexist in
one module — each is an independent render function. Compat mode is not like that: it
changes *module-level* output shape (babel rewrites the JSX expression in place; Vize
emits a standalone `render` export), so a module compiled half in compat mode and half
not would emit two mutually incompatible module shapes from one file. Compat is a
project-level decision, configured once, with no prologue form. This is documented in
`src/compat.rs`.

**Compat under Vapor is unsupported and diagnosed.** `@vue/babel-plugin-jsx` is a
vdom-era plugin; every output shape it defines is a `createVNode` tree, and it has no
Vapor equivalent. `jsxCompat: "babel"` with `jsxMode: "vapor"` therefore produces an
error naming both keys and the way out, rather than being silently ignored:

```
compiler.jsxCompat: "babel" is not supported with Vapor output: @vue/babel-plugin-jsx
has no Vapor equivalent. Use jsxMode "vdom" for babel compatibility, or drop jsxCompat
to use Vize's own Vapor semantics.
```

One diagnostic per render root, pinned by test.

**Unrecognized values fall back rather than erroring**, matching how an unrecognized
`jsxMode` string is already handled — a stray config value must never block a build.

## Why the mode is currently inert

No inventory row has been closed under it yet, and
`babel_compat_vdom_output_is_unchanged_for_now` asserts exactly that: compat and native
VDOM output are identical today. That test is the one the first behavioral PR flips —
when a divergence is closed, the two stop being equal and the assertion becomes the pin
for the new compat-only output. Landing the switch inert keeps each behavioral change a
small, reviewable diff against a stable surface.

## Verification

```
$ cargo test -p vize_atelier_jsx --test compat_mode
test result: ok. 5 passed; 0 failed
$ cargo test -p vize_atelier_jsx
29 test targets, all "test result: ok"
$ cargo test -p vize_carton --lib config
test result: ok. 60 passed; 0 failed
$ cargo clippy --workspace -- -D warnings -D clippy::wildcard_imports
Finished
$ vp run source:lengths
exit 0
```


ubugeeei:
Remaining blockers before this leaves draft, from the first CI run:

1. **`cargo-semver-checks (vize_carton)` fails** — `constructible_struct_adds_field`:
   `ConfigFeatureFlags.jsx_compat` is a new `pub` field on an externally-constructible
   `pub struct`, so existing struct literals would break. Needs a maintainer call:
   either `#[non_exhaustive]` on `ConfigFeatureFlags` (a one-time break that makes every
   future field addition non-breaking) or a version bump. `jsx_mode` was added to the
   same struct earlier, so there is precedent to follow — I did not want to pick the
   answer unilaterally.
2. **Plumbing not yet done**: JSON schema + generated types (`vp run gen:schema` /
   `gen:types`) and the Vite / unplugin / rspack option surfaces. `grep jsxMode` finds
   every sibling entry that still needs a `jsxCompat` counterpart.

Fixed in c4c873a: the wasm module's own `#[cfg(test)]` call sites for
`build_jsx_wasm_result` were not updated for the new `compat` argument. Worth noting for
anyone verifying locally — `cargo test -p vize_vitrine` reports `0 tests` for that
module on a host target, so the breakage only appears under `--all-targets` (which is
what CI runs).

ubugeeei:
## Unblocking the `cargo-semver-checks` decision

This PR is blocked on `constructible_struct_adds_field` for the new `pub ConfigFeatureFlags.jsx_compat`, with the choice framed as `#[non_exhaustive]` vs letting a version bump absorb it. Recording a recommendation so whoever picks this up is not blocked again.

**Recommendation: `#[non_exhaustive]` on `ConfigFeatureFlags`.**

Reasoning:

1. **This struct is going to keep growing.** It is the compiler feature-flag bag; `jsx_compat` is not the last flag it will ever gain. Absorbing the break with a version bump fixes it once and re-breaks on the next flag, every time, forever. `#[non_exhaustive]` fixes the class.
2. **It costs downstream almost nothing here.** `#[non_exhaustive]` prevents external struct-literal construction and exhaustive destructuring. For a flags bag that callers should be building via `Default::default()` and field updates, that is the intended usage anyway — and it is what makes adding a flag a non-event.
3. **`ubugeeei-redundancy.md` explicitly permits the alternative**, so this is not a policy question: "For the `0.x` major version era, do not distinguish meaningfully between minor and patch releases… users should assume every release may contain meaningful behavioral movement, compatibility changes, fixes, and new capability." A version bump would be *allowed*. It is just the worse of two allowed options, because it does not generalize.
4. **It is cheap to reverse before 1.0** if it turns out to constrain a real consumer.

If the maintainer prefers the version-bump route instead, that is entirely reasonable — it is one line either way, and this comment exists only so the PR does not sit blocked on an unanswered question.

## Separately, what this PR still needs before leaving draft

Recorded on #3391 but repeating here so it is in one place:

- the JSON schema + generated types (`vp run gen:schema` / `gen:types`);
- the Vite / unplugin / rspack option surfaces — `grep jsxMode` finds every sibling entry that needs a `jsxCompat` counterpart.

The mode itself is inert by design and proven so by `default_config_output_equals_explicit_native`, so there is no user-visible risk in it landing incomplete — but there is also no user-visible *value* until the plumbing is done, which is why it was correctly kept out of v0.306.0.

ubugeeei:
## Correcting my earlier recommendation — there is a third option, with direct precedent

My previous comment framed this as `#[non_exhaustive]` vs a version bump, and recommended `#[non_exhaustive]`. That framing was incomplete. A third option exists and this repo has already used it **on this exact struct**:

`69a44435d` (`feat!: add experimental RFC opt-ins`) added **five `pub` fields to `ConfigFeatureFlags`** under a `feat!:` marker. `.github/workflows/check.yml` L124-139 turns a `!:` or `BREAKING CHANGE:` marker in the PR title/body into `--release-type major` for `cargo semver-checks`:

```sh
if printf '%s\n' "$SEMVER_CHANGE_MARKER" | grep -Eq '^[[:alnum:]_-]+(\([^)]+\))?!:|^BREAKING CHANGE:'; then
  SEMVER_ARGS+=(--release-type major)
fi
```

So the gate is **already designed to be answered this way**, and adding a field to this struct is precisely the case it was used for before. Retitling this PR `feat(atelier)!: add an opt-in @vue/babel-plugin-jsx compatibility mode` makes the check pass with no API change at all.

**Revised recommendation: use the `!:` marker.** It matches existing precedent on the same struct, requires no permanent API constraint, and is the mechanism the CI gate already provides. `#[non_exhaustive]` remains defensible as a separate, deliberate API decision — but it should be made on its own merits in its own PR, not as a side effect of unblocking this one.

Credit where due: I did not find this; the closeout audit did, by looking for how the repo had solved it before instead of reasoning from first principles. That is the better instinct and I should have applied it first.

The remaining pre-draft work is unchanged: the JSON schema + generated types (`vp run gen:schema` / `gen:types`), and the Vite / unplugin / rspack option surfaces (`grep jsxMode` finds every sibling entry).

ubugeeei:
Rebased onto main and made green. Retitled with `!` — `ConfigFeatureFlags` gains a public `jsx_compat` field and `JsxCompat` is a new public type, which `cargo-semver-checks (vize_carton)` correctly flags as breaking.

Three source-length breaches the rebase surfaced, fixed by extraction rather than by trimming the change:

| file | before | after |
| --- | ---: | ---: |
| `vize_carton/src/config/loader.rs` | 360 (crossed 350) | 350 — `load_compiler_jsx_compat` moved to `loader/jsx.rs`, following the `loader/vapor.rs` sibling |
| `vize_carton/src/config/loader/tests.rs` | 418 (already over, grew from 379) | 341 — the six `jsxCompat` tests moved next to the loader they exercise |
| `vize_atelier_jsx/src/compile.rs` | 360 (crossed 350) | 262 — the inline test module moved to `compile/tests.rs`, with its two insta snapshots `git mv`d to the matching directory |

32 `vize_carton` + `vize_atelier_jsx` test binaries, workspace build, clippy and the source-length gate are green.

**Still draft for the reason the description gives**, and one of the two remaining items is blocked here rather than merely unfinished:

- **JSON schema + generated types.** `vp run gen:schema` shells out to `pkl eval`, and `pkl` is not installable in this environment. Without it `compiler.jsxCompat` is absent from `npm/cli/schemas/vize.config.schema.json`, so a `vize.config.json` setting the key would fail validation — which is exactly why the description holds the PR in draft. Needs a machine with `pkl`.
- **Vite / unplugin / rspack option surfaces.** Not blocked, just not done: `grep -rl jsxMode npm/builder` lists the sibling entries to mirror across `types.ts`, `compiler.ts`, `plugin/load.ts` and the rspack loader.

The switch itself is inert until those land — `JsxCompatMode::default()` is `Native` and `default_config_output_equals_explicit_native` pins that byte for byte — so the branch is safe to sit on main whenever the schema half is done.

Issue #3222, opened 2026-07-22
roadmap(canon): win the vue-tsc parity scorecard


ubugeeei:
Tracks the typechecker product workstream (parent: #2971 evidence campaign; peer of the #3131 tree).

## Outcome

`vize check` beats `vue-tsc` on correctness parity, span parity, latency, and memory — with machine-checked evidence, not claims.

## Scorecard dimensions (each becomes a checked, CI-enforced gate)

- [x] Diagnostic parity: whole-ecosystem FP/FN divergence ledger (`tools/fixtures/typecheck-divergence.mjs`, schema `vize.fixtureTypecheckDivergence`) promoted from weekly-only (`real-project-matrix.yml`) to a per-PR fast-tier ratchet with a checked-in baseline that PRs may only hold or improve.
- [x] Span parity: prop-type errors anchored at the attribute name byte-for-byte with vue-tsc (divergence 2 of the Tier S report; change point `crates/vize_canon/src/virtual_ts/expressions/component_props.rs`), plus exact-range oracle sides for emit payload, `v-model`, CSS module, JSX/TSX, Vapor, and generic-component errors.
- [x] Performance: cold/warm/single-leaf/shared-edit latency and peak RSS budgets ENFORCED (today report-only in `tests/performance/support/incremental-metrics.ts`) on Misskey and Vue Vben Admin; regression trips `vue-parity`.
- [x] Robustness: churn/leak/cancellation stress (repeated edits + create/delete churn, bounded RSS growth, no leaked Corsa sessions, byte-stable diagnostics).
- [x] Legacy: Vue 2.7 / class-component LSP broken/repaired oracles; `vue2-elm` exercised.

## Delivery

Small conventional PRs, one gate per PR, auto-merge after green. Evidence lands as CI artifacts with commit/runtime/machine/fileCount/timing.

ubugeeei:
Audited the first two scorecard dimensions. Dimension 1 has a gap worth acting on: **the ratchet is written but not enforced on any PR.**

## Dimension 1 — the ratchet exists, the "per-PR" half does not

`tests/tooling/compat-ratchet.test.ts` is exactly the gate this dimension asks for. Its rules are hold-or-improve over a checked-in baseline:

```
falsePositiveCount   <=
falsePositiveRatio   <=
falseNegativeCount   <=
falseNegativeRatio   <=
sharedCount          >=
baselineDiagnosticCount ==
```

[...]

PR #3443, opened 2026-07-30
fix(canon): type-check every native element v-bind, not just boolean-ish names

ubugeeei:
Part of the #3222 parity scorecard (dimension 1: diagnostic parity). Found by
running vue-tsc 3.3.4 as an oracle over synthetic and hydrated-fixture corpora.

## Divergence — a silent false negative class

A statically named `v-bind` on a native element was only checked against the
element's own prop type when the attribute name appeared in a hard-coded
boolean-ish list (`disabled`, `checked`, `hidden`, `multiple`, ...). Every other
native prop-type error was silently dropped.

Minimal reproduction, `src/App.vue`:

```vue
<script setup lang="ts">
const num = 1
</script>

<template>
  <a :href="num">link</a>
</template>
```

| | output |
| --- | --- |
| vue-tsc 3.3.4 | `src/App.vue(7,7): error TS2322: Type 'number' is not assignable to type 'string'.` |
| vize check (before) | *(nothing)* |
| vize check (after) | `src/App.vue(7,7): error TS2322: Type 'number' is not assignable to type 'string'.` |

Over a 21-binding probe file (`:href`, `:src`, `:alt`, `:maxlength`,
`:placeholder`, `:type`, `:colspan`, `:id`, `:title`, `:tabindex`, `:for`,
`:role`, `:multiple`, plus `:class`/`:style`/`:key`/`:data-*`/`:aria-*`),
vue-tsc reported 11 diagnostics and vize reported 1. After the fix both report
the same 11, at the same positions, with the same messages — exact parity.

## Fix

The name list is removed rather than extended. An attribute the element type
does not declare (`:data-id`, `:aria-label`, a custom attribute) already
resolves through `__VizeNativeElementProp<..., Prop>` to `unknown`, which
accepts any value, so the permissive attributes stay permissive with no list to
maintain. Dynamic names (`:[key]="x"`) and `v-bind="object"` are still skipped —
neither names a statically known prop.

## Evidence that this does not add false positives

Two pinned binaries (pre-fix and post-fix, copied out of the build directory so
a concurrent build could not swap them) were run over the same inputs.

**Adversarial clean-code probe** — 33 correct native bindings including
`:class`/`:style` as object, array and string, `:key` in `v-for`, `:data-*`,
`:aria-*`, SVG (`:viewBox`, `:cx`), `.prop`/`.attr` modifiers, a dynamic
`:[str]`, a custom element, and `any`-typed values. Both tools report **zero**
diagnostics, before and after.

**Hydrated ecosystem slices** — 7 large slices of the pinned fixtures
(misskey `components/global`, element-plus `docs/examples/{button,input,table}`,
vitepress `theme-default`, vue-element-admin `src/components` + dashboard,
vue-vben-admin `playground/src/views`, nuxt-ui `content/examples`, pinia docs
theme), classified with the ledger comparator
(`tools/fixtures/typecheck-divergence.mjs`). 1 587 vize diagnostics vs 1 590
vue-tsc diagnostics. The complete before/after diff is:

```
< misskey: vize=128 vue-tsc=123 shared=111 FP=17 FN=12
> misskey: vize=127 vue-tsc=123 shared=111 FP=16 FN=12
<    +FP .../MkPageHeader.tabs.vue(11,11) TS7006: Parameter 'el' implicitly has an 'any' type.
```

One false positive **removed** (the native binding now supplies a contextual
type for an inline arrow), zero added, nothing else changed.

**Compat ratchet** — `tests/tooling/compat-ratchet.test.ts` over all 9 hydrated
probes: 12/12 pass, every probe still 0 FP / 0 FN. The checked-in baseline is
untouched.

**Codegen cost** — the generated virtual TS for the nuxt-ui examples slice
(320 virtual files) grows from 1 730 831 to 1 743 185 bytes, **+0.71 %**.
Wall-clock was not measured: this machine is running many concurrent builds and
repeated runs of the *same* binary varied by more than 2x, so no timing number
here would be honest.

## Tests

- `native_prop_check_covers_every_static_attribute_name` (new): asserts `:href`
  and `:tabindex` on `<a>` each emit a check against that element's own prop
  type. Fails pre-fix — the old allowlist emitted neither.
- `native_prop_check_ignores_dynamic_names_and_components` (was
  `..._ignores_non_boolean_names_dynamic_names_and_components`): keeps pinning
  the two cases that must stay unchecked, minus the assertion that encoded the
  defect.
- `tests/_fixtures/_projects/typecheck-errors` gains `<a :href="count">` (an
  intentional error) and `<span :data-count="count">` (which must stay quiet).
  The `typecheck-errors` snapshot now records
  `error:17:9 [TS2322] Type 'number' is not assignable to type 'string'.` and
  `errorCount` 25 -> 26. vue-tsc over the identical file reports exactly the
  same three diagnostics — verified with the oracle harness.

### Verified failing before the fix

With the pre-fix binary and the new fixtures/snapshots in place,
`node --test` over the `test:check:fixtures` list goes from 78 pass / 4 fail to
74 pass / 8 fail. The four extra failures are exactly the four snapshots that
encode the new behavior:

```
not ok 2  - compiler-macros check (type checker)
not ok 9  - ecosystem-products check (type checker)
not ok 18 - style-preprocessors check (type checker)
not ok 20 - typecheck-errors check (type checker)
```

with `typecheck-errors` reporting `errorCount=25` and the missing
`-   "error:17:9 [TS2322] Type 'number' is not assignable to type 'string'."`.

The 4 failures common to both binaries are `javascript-sfc-checkjs-oracle.ts`,
which fails identically on unmodified `main` in this environment and is
unrelated.

## Snapshot churn

Ten `insta` snapshots and three `--show-virtual-ts` JSON snapshots record the
extra check statements. Their **diagnostics are unchanged** — only the generated
TypeScript differs; `errorCount` moves in `typecheck-errors` alone, and only by
the one intentional error added to the fixture.

## Gates

- `cargo test -p vize_canon` — all suites pass
- `cargo clippy --workspace -- -D warnings -D clippy::wildcard_imports` — clean
- `rustfmt --edition 2024 --config style_edition=2024 --check` on every changed
  Rust file — clean
- `crates/vize_canon/src/virtual_ts/tests.rs` held at exactly its base length
  (2220) so the over-budget-file rule still passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved template type checking for statically named native element `v-bind` bindings (e.g., `href`, `tabindex`) to validate against the element’s full prop types.
  * Avoids false diagnostics when Vue typings omit `NativeElements`.
  * Preserves existing behavior for dynamic `v-bind` arguments and component/dynamic attributes.
  * Refined generated virtual TS output and improved template expression source mapping selection.
* **Tests**
  * Added a regression test covering the “missing `NativeElements`” Vue typings scenario.
  * Updated SFC and virtual TS assertions to match revised output formatting.


ubugeeei:
## Update — CI found a real false positive; fixed, with a test

The first push was red on `clippy-and-test`:
`check_generic_script_setup_resolves_type_referencing_generic_param` reported a
diagnostic that did not exist before:

```
error:12:41 [TS2694] Namespace '".../node_modules/vue/index"' has no exported member 'NativeElements'.
```

That is **not a stale test**. Widening the check also widened the reach of the
`import('vue').NativeElements` reference, which the generated code named at
every use site. Any project whose Vue typings do not export it — Vue 2.7, an
older Vue 3 minor, a trimmed or shimmed `vue`, a workspace resolving `vue` to a
stub — would get one spurious `TS2694` per authored native attribute. The
adversarial probe and all seven hydrated slices ship a full modern Vue, so none
of them could see it.

Reproduced outside the test suite, with a `vue` whose `index.d.ts` predates
`NativeElements` and a two-binding template:

```
"diagnostics": [
  "error:6:16 [TS2694] Namespace '.../node_modules/vue/index' has no exported member 'NativeElements'.",
  "error:7:16 [TS2694] Namespace '.../node_modules/vue/index' has no exported member 'NativeElements'."
],
"errorCount": 2
```

### Fix

The module is now named exactly once per file, behind an alias:

```ts
type __VizeNativeElements = import('vue').NativeElements;
type __VizeNativeElement<Tag extends PropertyKey> = Tag extends keyof __VizeNativeElements ? __VizeNativeElements[Tag] : unknown;
```

A missing export makes that single alias an error type, which TypeScript
propagates as `any`, so every check below it accepts any value and the file
degrades to **unchecked** — the same permissive outcome an undeclared attribute
already gets. Absence of the type can no longer mean "error".

Same workspace, after the fix:

```
"diagnostics": [],
"errorCount": 0
```

### Tests added for exactly this shape

- `native_prop_checks_stay_silent_when_vue_has_no_native_elements`
  (`type_checker/tests/template_block.rs`): builds a project whose
  `node_modules/vue` exports no `NativeElements`, checks a template with two
  native bindings, and asserts the **full** diagnostic list is empty.
- `native_prop_check_uses_vue_jsx_type_and_authored_subspans` now asserts
  `output.code.matches("import('vue').NativeElements").count() == 1`, so a
  future edit cannot reintroduce a per-use-site reference.

`cargo test -p vize --test check_cli` is back to 45 passed / 0 failed.

### Everything else re-verified with the new binary

- `native-wide` (11 bindings), `native-clean` (33 correct bindings),
  `native-testcfg`, and the updated `typecheck-errors` fixture: **exact parity**
  with vue-tsc in every case.
- Wide ecosystem probe: byte-identical to the numbers reported above — misskey
  still 17 -> 16 false positives, nothing else changed.
- `tests/tooling/compat-ratchet.test.ts`: all 9 probes 0 FP / 0 FN, baseline
  untouched.
- `cargo fmt --all --check` (toolchain rustfmt inside `nix develop`), and
  `cargo clippy --workspace -- -D warnings -D clippy::wildcard_imports`: clean.

Rebased on current `origin/main`.

ubugeeei:
## Update 2 — root cause of the `NativeElements` resolution failures, and the fix

Recording the analysis in full, because it is the expensive part of this PR.

### What the generated code does

A native prop check needs Vue's element table:

```ts
const __vize_native_prop_check_13: __VizeNativeElementProp<__VizeNativeElement<"a">, "href"> = (num);
```

`__VizeNativeElement` bottoms out at `import('vue').NativeElements`. **Where that
module reference is written decides whether an unresolvable Vue is silent or
loud**, and that is the whole story of the two CI failures.

### Failure 1 — `crates/vize/tests/check_cli.rs` (fixed in update 1)

The reference was written **at every use site**. A fixture whose `vue` typings
omit `NativeElements` got one `TS2694` per authored attribute:

```
error:12:41 [TS2694] Namespace '".../node_modules/vue/index"' has no exported member 'NativeElements'.
```

Real reach: Vue 2.7, an older Vue 3 minor, a trimmed or shimmed `vue`, a
workspace resolving `vue` to a stub. Update 1 collapsed the reference to a single
per-file alias, so the failure became one error type that TypeScript propagates
as `any`.

### Failure 2 — `vize_maestro` `editor_relative_vue_imports_resolve_existing_siblings`

Update 1 was not enough. That fixture has **no `vue` package at all**, so the
alias itself failed to resolve — and because it was emitted *inside the module's
template closure*, it sat among mapped regions and the diagnostic was attributed
to user source:

```
Diagnostic { range: 28:25..28:39, message: "Cannot find module 'vue' or its corresponding type declarations." }
```

So the same defect, one level up: an unresolvable Vue must never produce a
diagnostic on authored bytes.

### The fix

All three native-element aliases move into the shared modern helper text
(`vue_type_aliases_text!` in `virtual_ts/helpers.rs`), which the batch path emits
**once per program** into `__vize_helpers.d.ts`:

```ts
type __VizeNativeElements = import('vue').NativeElements;
type __VizeNativeElement<Tag extends PropertyKey> = Tag extends keyof __VizeNativeElements ? __VizeNativeElements[Tag] : unknown;
type __VizeNativeElementProp<Element, Prop extends PropertyKey> = Prop extends keyof Element ? Element[Prop] : unknown;
```

That is exactly where the pre-existing `type __Ref<T> = import('vue').Ref<T>;`
already lives, and it has always tolerated a missing Vue for the same reason:
the helpers file carries no mapping back to authored source, so an unresolved
reference cannot be attributed to user code, and the alias degrades to `any`,
leaving every check permissive. `generate_native_prop_helpers` and its per-file
emission are deleted.

The legacy Vue 2 path is untouched — it uses `LEGACY_VUE_TYPE_HELPERS`, a
separate constant that never names `NativeElements`, which
`legacy_vue2_does_not_require_vue_native_elements` still pins.

### This is also the expected fix for `pr-benchmark-budget`

The budget failed on the previous pushes (`Type check (1T)` +6.5%, then +50%).
Declaring the conditional-type helpers **per file** meant TypeScript
re-instantiated `keyof __VizeNativeElements` and the per-element `keyof` for
every generated module instead of reusing a program-wide cache. Hoisting them to
one declaration per program is precisely the shape that lets the instantiations
be shared. Whether that is enough is for CI to answer — that number is not
something this machine can measure honestly (repeated runs of the *same* binary
varied by more than 2x under concurrent load).

### Verified with the hoisted build

Pinned binary, same corpora as before:

- `native-wide` (11 bindings), `native-clean` (33 correct bindings), and the
  updated `typecheck-errors` fixture: **exact parity** with vue-tsc 3.3.4.
- A workspace whose `vue/index.d.ts` predates `NativeElements`, two native
  bindings: `"diagnostics": []`, `errorCount: 0`.
- `cargo test -p vize_maestro --lib`: 590 passed, 0 failed.
- `tests/tooling/compat-ratchet.test.ts`: 9/9 probes, 0 FP / 0 FN, baseline
  untouched.
- `cargo fmt --all --check` and
  `cargo clippy --workspace -- -D warnings -D clippy::wildcard_imports`: clean.
- `crates/vize_canon/src/virtual_ts/helpers.rs` held at its base length (598) so
  the over-budget-file rule still passes.

### One local result I could not confirm and do not believe

`crates/vize/tests/check_cli.rs` `check_can_emit_declarations` and
`check_declaration_emit_uses_tsconfig_options` failed locally with
`Result::unwrap()` on `Error("EOF while parsing a value", line: 1, column: 0)` —
empty stdout from `CARGO_BIN_EXE_vize`. Running the *same* CLI invocation by hand
against a pinned binary produces correct output and `EXIT=0`, byte-identical
between the pre-change and post-change builds, and the same suite passed 45/45
earlier in this session. This lane shares a `CARGO_TARGET_DIR` with other agents,
so the test executed a binary another build was mid-write on. CI has its own
clean target dir and is the authority here.

ubugeeei:
## Update 3 — declaration emit, and the answer to "is `NativeElements` guaranteed available everywhere?"

### Why declaration emit moved, and why it should not have

`__vize_helpers.d.ts` is itself an **emitted declaration artifact**, and update 2
put the native-element aliases into `vue_type_aliases_text!`, which
`DECLARATION_HELPERS_DTS` splices. So the shipped helpers file grew three
aliases, including `import('vue').NativeElements`.

The snapshot diff was exactly that and nothing else — the emitted `.d.ts` for the
user's own SFC was byte-identical — but blessing it would have been wrong.
`DECLARATION_HELPERS_DTS` already documents the rule it broke:

> Deliberately excludes the `vite/client` reference, the `ImportMeta`
> augmentation, and the value-level macro declarations: emitted declarations
> never reference those, **and they must not leak into consumer programs**.

An emitted declaration never contains a native prop check, so the alias is dead
weight there — and unlike the internal virtual project, where an unresolvable
alias is harmlessly `any`, a broken alias in a *shipped* `.d.ts` surfaces in the
**consumer's** build. Adding it would have pushed vize's Vue-version floor onto
every downstream package.

The aliases now live only in `vue_type_helpers_text!`, which feeds the
type-checking preamble (`VUE_TYPE_HELPERS` and `SHARED_PREAMBLE_DTS`) and not
`DECLARATION_HELPERS_DTS`. `cargo test -p vize --test build_declaration_emit_cli`
is 4/4 with the original snapshot restored — no snapshot was accepted.

### Is `NativeElements` resolution guaranteed on every path that checks a native binding?

No, and it must not need to be. That was the wrong shape to aim for; the
invariant is instead:

> the module is named in exactly one place, and that place carries no mapping
> back to authored source.

Both halves now hold, and they are what make absence permissive rather than
loud:

1. **One name.** `import('vue').NativeElements` appears once, in
   `__VizeNativeElements`. `native_prop_check_uses_vue_jsx_type_and_authored_subspans`
   asserts `matches(...).count() == 1`, so a future edit cannot reintroduce a
   per-use-site reference.
2. **Unmapped.** It sits in the shared helper text next to the pre-existing
   `type __Ref<T> = import('vue').Ref<T>;`, which has always tolerated a missing
   Vue for the same reason.

Three paths reach a native binding, and all three are covered:

| path | preamble | evidence |
| --- | --- | --- |
| batch / `vize check` | hoisted into `__vize_helpers.d.ts` | `native_prop_checks_stay_silent_when_vue_has_no_native_elements`; a hand-built workspace whose `vue/index.d.ts` predates `NativeElements` reports `"diagnostics": []`, `errorCount: 0` |
| LSP / maestro | inline module preamble | `editor_relative_vue_imports_resolve_existing_siblings` (fixture with no `vue` at all) passes; `cargo test -p vize_maestro --lib` 590/590 |
| declaration emit | not emitted at all | the alias never reaches `DECLARATION_HELPERS_DTS` |

Legacy Vue 2 never names it: that path uses `LEGACY_VUE_TYPE_HELPERS`, pinned by
`legacy_vue2_does_not_require_vue_native_elements`.

### Where the previous run stood

`fmt-rust`, `vue-parity` and `pr-benchmark-budget` all **passed** on the update-2
push. The benchmark result is worth recording: it went from +50.34% /
+47.78% (`Type check 1T` / `max`) with per-file helper declarations to within
budget once the helpers were hoisted to one declaration per program — the
conditional types were being re-instantiated for every generated module instead
of hitting TypeScript's instantiation cache.

`clippy-and-test` failed only on the declaration snapshot addressed above.

`test-scripts` failed on an LSP typecheck timeout —
`corsa request outran the 10000ms bridge bound; abandoned it` on a single small
template, preceded by `failed to release corsa snapshot ...: release queue is
closed`. I have not tied that to this change and it does not reproduce locally
(`cargo test -p vize_maestro --lib` 590/590). Treating it as unexplained rather
than as a flake until this run says otherwise.

### Local result I still do not believe

`check_can_emit_declarations` and `check_declaration_emit_uses_tsconfig_options`
keep failing *locally* with `Error("EOF while parsing a value", line: 1, column: 0)`
— empty stdout from `CARGO_BIN_EXE_vize`. Both passed in CI on the previous run,
and running the same CLI invocation by hand against a pinned binary produces
correct output and `EXIT=0`. This lane shares a `CARGO_TARGET_DIR` with other
agents, so the test is executing a binary another build is mid-write on.

ubugeeei:
## Correction to update 3

Update 3 on this PR attributed the two declaration-emit failures — `check_can_emit_declarations` and `check_declaration_emit_uses_tsconfig_options` — to shared-build-directory contamination.

**That was wrong. CI has since reproduced both on a clean checkout, and they are the sole remaining blocker on this PR.** Recording the correction here so the PR history does not carry a false all-clear.

For whoever picks this up, the working hypothesis (not yet verified):

- The last change removed the native-element aliases from `DECLARATION_HELPERS_DTS`, whose doc states that emitted declarations must not leak helpers into consumer programs.
- But the generated virtual TypeScript references `__VizeNativeElement*`, so the declaration-emit **program** appears to still need those aliases even though the emitted `.d.ts` must not **ship** them.
- Those are two different concerns that were conflated. `crates/vize_canon/src/batch/virtual_project/snapshots/…__pruned_helpers.snap` moving earlier in the run suggests there is a helper-pruning step tied to that path, which is probably where the distinction belongs.

Everything else that CI drove out of this change is fixed and documented above: the per-binding `TS2694` when Vue's typings lack `NativeElements`, the same error resurfacing on user code through the LSP path, and a +50% type-check benchmark regression caused by declaring the helper conditional types per file (which defeated TypeScript's instantiation cache — hoisting them to a single program-wide declaration fixed it). `pr-benchmark-budget`, `vue-parity`, `test-scripts` and `fmt-rust` all pass.

This PR was deliberately **excluded from the v0.305.0 release**. Turning permissive-by-default into error-by-default across every native element binding is not a change to ship half-verified, and it lost nothing by waiting — the fix itself is sound (10 of 11 native DOM prop-type errors were silently dropped; removing the hard-coded name list gives byte-identical span parity with vue-tsc at 11/11, removes a pre-existing false positive on misskey, adds zero false positives across seven hydrated ecosystem slices, and keeps the compat ratchet at 0 FP / 0 FN).

ubugeeei:
## Update 4 — the two declaration-emit failures are fixed

`606f7dfbb` fixes `check_can_emit_declarations` and
`check_declaration_emit_uses_tsconfig_options`, the two remaining blockers.

### Root cause — not the one the previous update guessed

The correction on this PR proposed that `DECLARATION_HELPERS_DTS` had wrongly
dropped the native-element aliases and that the declaration-emit *program* still
needed them. That is not what is happening, and `DECLARATION_HELPERS_DTS` is
correct as it stands: an emitted `.vue.d.ts` never contains a native prop check
(the check is a `const` inside the render closure, which declaration emit
discards), so shipping the aliases would only push vize's Vue floor onto
consumers for nothing.

The failure is in the *virtual project's* `__vize_helpers.d.ts` — the shared
ambient preamble, a different file that happens to share the name. The `vue` that
`crates/vize/tests/check_cli.rs` writes for its CLI projects
(`write_test_vue_runtime_dom_stub`) does not export `NativeElements`, so:

```
Error: corsa error (exit code 1): __vize_helpers.d.ts(44,43): error TS2694:
  Namespace '".../node_modules/vue/index"' has no exported member 'NativeElements'.
```

Reproduced locally byte-for-byte, on this branch, before any change:

```
$ cargo test -p vize --test check_cli -- declaration
test check_can_emit_declarations ... FAILED
test check_declaration_emit_uses_tsconfig_options ... FAILED
  left: Some(1)
 right: Some(0)
```

The degrade this PR designed for was half-implemented. A `vue` without
`NativeElements` does make the alias an error type, which propagates as `any` and
leaves every check permissive — that part works. But the alias *also* emits a real
`TS2694`. The diagnostics path drops it, because the helpers text carries no
mapping back to authored source and an unmapped diagnostic can never be attributed
to user code — which is why `native_prop_checks_stay_silent_when_vue_has_no_native_elements`
passes and the degrade looked complete. Declaration emit does not go through that
path: `BatchExecutor::emit_declarations` shells out to Corsa and treats any
non-zero exit as fatal, so the same unmapped diagnostic killed the whole emit.

One diagnostic, two consumers, only one of them tolerant.

### Fix

`// @ts-ignore` on the single line that names the module — the same idiom the
generated setup scope already uses for `TS6196`. The alias still resolves to an
error type, so the checks stay permissive; nothing is reported on any path.

Behaviour with a real `vue` is unchanged. Probed directly against
tsgo `7.0.0-dev.20260602.1` and `vue@3.6.0-beta.10`, same helper text with and
without the guard:

```
=== real vue, helpers WITHOUT guard ===
src/App.vue.ts(3,7): error TS2322: Type 'number' is not assignable to type 'string'.
=== real vue, helpers WITH @ts-ignore ===
src/App.vue.ts(3,7): error TS2322: Type 'number' is not assignable to type 'string'.
```

and against a `vue` that stops short of `NativeElements`:

```
=== stub vue, helpers WITHOUT guard ===
__vize_helpers.d.ts(2,43): error TS2694: ... has no exported member 'NativeElements'.
exit=2
=== stub vue, helpers WITH @ts-ignore ===
exit=0
```

### Tests

Two, both failing on the pre-fix code:

- `crates/vize/tests/check_declaration_emit_vue_floor_cli.rs` —
  `declaration_emit_survives_a_vue_without_native_elements`. Writes its own
  `node_modules` with a `vue` that predates `NativeElements` (rather than linking
  the workspace's), runs `vize check --declaration`, and pins exit `0`, zero
  diagnostics, the exact three emitted declaration paths, and that neither the
  emitted `App.vue.d.ts` nor the emitted helper file ships `__VizeNativeElement*`.
  Pre-fix it reproduces the CI error exactly:

  ```
  Error: corsa error (exit code 1): __vize_helpers.d.ts(44,43): error TS2694: ...
    left: Some(1)
   right: Some(0)
  ```

- `native_element_alias_is_guarded_at_its_only_declaration_site` — pins the guard
  line immediately before the alias, in `SHARED_PREAMBLE_DTS` and in the per-file
  `VUE_TYPE_HELPERS` alike, and that the module is still named exactly once.

The `vue` stub in `check_cli.rs` is deliberately left without `NativeElements`, so
`check_can_emit_declarations` and `check_declaration_emit_uses_tsconfig_options`
keep guarding this too.

### Verification

```
cargo test -p vize --test check_cli -- declaration      3 passed; 0 failed
cargo test -p vize --test check_declaration_emit_vue_floor_cli   1 passed; 0 failed
cargo test -p vize --test check_declaration_emit_cli    2 passed; 0 failed
cargo test -p vize_canon                                every suite ok (655 + 22 suites)
cargo clippy --workspace -- -D warnings -D clippy::wildcard_imports   clean
rustfmt 1.9.0 --edition 2024 --config style_edition=2024 --check      clean
```

Snapshot churn: the helpers text gains one comment line, so 37 insta snapshots
record it plus the resulting `@vize-map` byte-offset shift (+165). No diagnostic
changes in any of them. Files touched stay inside the 350-line budget
(`helpers.rs` is unchanged at 598 lines — the comment goes inside the existing
string literal).

ubugeeei:
## Update 5 — the third downstream failure, and the fix that covers all of them

The `@ts-ignore` in update 4 fixed the two declaration-emit failures and surfaced a
third, exactly the way this PR has surfaced them one at a time.
`check_server_maps_dependency_patch_diagnostics_to_the_parent_sfc` went red:

```
  left: Array [
    {"code": "TS6196", "line": 47, "column": 6, "severity": "hint",
     "message": "'__VizeNativeElement' is declared but never used."},
    {"code": "TS6196", "line": 48, "column": 6, "severity": "hint",
     "message": "'__VizeNativeElementProp' is declared but never used."}]
 right: Array []
```

Reproduced locally. `line: 47` / `column: 6` are raw positions inside
`__vize_helpers.d.ts`; the request was for an 8-line `src/App.vue`.

The two conditional aliases are named only by the native prop checks, so a program
whose files bind no native prop leaves them unused. The `TS2694` on the alias above
them was previously masking that — suppress the error and the unused report appears.

`@ts-ignore` does **not** cover this one. It filters errors; these arrive on the
suggestion channel. Verified rather than assumed: putting all three aliases on one
line behind one `@ts-ignore` still reported both hints, just at the packed line.

The fix is to keep the aliases referenced:

```ts
declare function __vizeNativeElementProp<__Tag extends PropertyKey, __Prop extends PropertyKey>(value: __VizeNativeElementProp<__VizeNativeElement<__Tag>, __Prop>): void;
```

Ambient declarations are exempt from unused reporting themselves — which is exactly
how `__vForList` already keeps `__VForEntry` and `__VizeIsAny` alive in the same
file — so this closes the loop without adding a diagnostic of its own. It is never
called. The `// @ts-ignore` stays for `TS2694`/`TS2307`, which it does cover.

That the hints reached the client at all is a separate defect: unmapped diagnostics
from the generated program are dropped when they are errors and not when they are
hints. Filed as #3478 rather than widened into this PR.

### Verification

```
cargo test -p vize --test check_server_cli                       1 passed; 0 failed
cargo test -p vize --test check_cli -- declaration               3 passed; 0 failed
cargo test -p vize --test check_declaration_emit_vue_floor_cli   1 passed; 0 failed
cargo test -p vize --test check_declaration_emit_cli             2 passed; 0 failed
cargo test -p vize_canon --lib                                   655 passed; 0 failed
rustfmt 1.9.0 --edition 2024 --config style_edition=2024 --check  clean
```

All three tests that this PR drove red — `check_can_emit_declarations`,
`check_declaration_emit_uses_tsconfig_options` and
`check_server_maps_dependency_patch_diagnostics_to_the_parent_sfc` — now pass, and
their harness `vue` stubs are deliberately left without `NativeElements` so they
keep guarding this.

`native_element_table_is_guarded_and_kept_referenced` (renamed from the update-4
test) now pins both halves: the guard line immediately before the alias, and the
ambient reference, in `SHARED_PREAMBLE_DTS` and the per-file `VUE_TYPE_HELPERS`
alike. It fails on the pre-fix code in both directions.

PR #3472, opened 2026-07-31
fix(canon): use tsc's real default `exclude` instead of the package folders

### LINKED ISSUE #3395: fix(canon): the default `exclude` drops literal `node_modules` include segments that tsc keeps
Found while fixing #3385. Filing separately because it is the opposite-direction half of the same modelling mistake and it *adds* program roots rather than removing them.

## The divergence

Vize models the documented default `exclude` as three literal glob specs anchored at the tsconfig directory (`default_exclude_specs` in `crates/vize/src/commands/check/tsconfig_inputs/glob.rs`). `tsc` does not have that default at all.

In TypeScript, `node_modules` / `bower_components` / `jspm_packages` are excluded by `implicitExcludePathRegexPattern`, which is baked into how a wildcard `include` segment is compiled — that part is what #3385 fixes. The *`exclude` default itself* is only `[outDir, declarationDir]`. So a package folder named by a **literal** include segment is a program root in `tsc`, and vize's anchored exclude spec drops it.

## Reproduction

Tree (under a `mktemp -d`, so no `.gitignore` hides anything):

```
node_modules/top/index.ts
packages/a/index.ts
```

```json
{ "compilerOptions": { "noEmit": true, "types": [] }, "include": ["node_modules/**/*.ts"] }
```

```
$ tsgo -p tsconfig.json --listFiles
node_modules/top/index.ts
```

Expected: `node_modules/top/index.ts` is a program root — the first segment is literal, so nothing implicitly excludes it, and the default `exclude` does not contain `node_modules`.

Actual (vize): collected as empty, because `default_exclude_specs` contributes a `node_modules/**/*` exclude. Writing `"exclude": []` explicitly is currently the only way to get vize to agree, which is itself a tell: in `tsc` the two spellings list the same files.

## Reachability

Deliberate, not accidental: a project has to name a package folder in `include`. Vendored-dependency and "typecheck one dependency" setups do this. Low frequency, but it silently checks nothing rather than producing a wrong diagnostic, which is the harder failure to notice.

## Suggested direction

Drop `default_exclude_specs` now that #3385 applies the implicit exclusion where `tsc` applies it (`tsconfig_inputs/implicit_exclude.rs`), and replace it with the real default — `outDir` and `declarationDir`, which vize does not exclude by default today either. `TsconfigDeclarationOptions` already resolves both, so the plumbing exists.

Worth pairing with a compat-ratchet run: unlike #3385 this direction can only add program roots, so diagnostic counts on the probes can move up.

ubugeeei:
## The divergence

`vize check` modelled the default `exclude` as three literal globs — `node_modules`,
`bower_components`, `jspm_packages` — anchored at the tsconfig directory
(`default_exclude_specs`). `tsc` has no such default.

Those three names come from `implicitExcludePathRegexPattern`, which is compiled *into* a
wildcard `include` segment (that half is #3401 / `tsconfig_inputs::implicit_exclude`). The
`exclude` field's own default is `[outDir, declarationDir]`. Modelling it as an anchored
`exclude` was therefore wrong in both directions:

* an include entry naming a package folder **literally** is a program root in `tsc`, and
  vize dropped it — the `exclude` shadowed the literal segment;
* `outDir` / `declarationDir` were never excluded, so build output became a program root
  whenever a project left `exclude` unset.

## Reproduction

Under `mktemp -d` outside any repository (a checkout's `.gitignore` hides `node_modules`
and the condition with it):

```
node_modules/top/index.ts
packages/a/index.ts
```

```json
{ "compilerOptions": { "noEmit": true, "types": [] }, "include": ["node_modules/**/*.ts"] }
```

| | `tsgo -p tsconfig.json --listFiles` | vize before | vize after |
| --- | --- | --- | --- |
| `include: ["node_modules/**/*.ts"]` | `node_modules/top/index.ts` | *(nothing)* | `node_modules/top/index.ts` |
| same + `"exclude": []` | `node_modules/top/index.ts` | `node_modules/top/index.ts` | `node_modules/top/index.ts` |
| `include: ["bower_components/**/*.ts", "jspm_packages/**/*.ts"]` | both files | *(nothing)* | both files |
| `outDir: "dist"`, no `exclude` | `src/index.ts`, `types/built.ts` | + `dist/built.ts` | `src/index.ts`, `types/built.ts` |
| `declarationDir: "types"`, no `exclude` | `dist/built.ts`, `src/index.ts` | + `types/built.ts` | `dist/built.ts`, `src/index.ts` |
| `outDir: "dist"` + `"exclude": []` | all three | all three | all three |

That the two spellings `exclude` absent / `"exclude": []` disagreed in vize but not in
`tsc` was the tell: in `tsc` they list the same files whenever no output directory is
configured.

## The fix

`TsconfigInputSpec` now carries `outDir` / `declarationDir` as globs over the directories
they resolve to, and `TsconfigInputSpec::effective_excludes` returns the user's `exclude`
when the config chain declares one — including `[]` — and those two otherwise.
`default_exclude_specs` is gone.

Details pinned by tests, each taken from `tsgo`:

* the spec is anchored at the **resolved** directory, so an absolute `outDir` excludes the
  same tree a relative one does;
* both are ordinary compiler options, so a config that `extends` a base declaring `outDir`
  inherits the exclusion;
* an empty string is not a directory and yields no spec (`tsc`'s default-exclude check is a
  truthiness test on the option);
* `files` entries still bypass `exclude` entirely.

## Tests

* `crates/vize/src/commands/check/tsconfig_inputs/default_exclude_tests.rs` — 9 new cases
  for the `outDir` / `declarationDir` default.
* `crates/vize/src/commands/check/tsconfig_inputs/package_folder_tests.rs` — 2 new cases
  for literal package-folder include segments, and
  `a_recursive_wildcard_below_a_literal_package_folder_segment_stays_shallow` loses the
  `"exclude": []` it only needed to work around this defect.

Case directories live under `std::env::temp_dir()` for the reason #3401 established: inside
the checkout, `.gitignore` hides `node_modules` and the assertions pass without the fix.

**Failing before the fix.** Restoring the five pre-fix collector sources under the new tests
(`git checkout origin/main -- tsconfig_inputs.rs glob.rs spec.rs loader.rs collect.rs`,
re-registering the test modules):

```
test result: FAILED. 10 passed; 8 failed
    default_exclude_tests::an_absolute_out_dir_is_excluded
    default_exclude_tests::an_out_dir_inherited_through_extends_is_still_excluded
    default_exclude_tests::declaration_dir_is_excluded_when_the_config_declares_no_exclude
    default_exclude_tests::out_dir_beats_an_include_that_names_it
    default_exclude_tests::out_dir_is_excluded_when_the_config_declares_no_exclude
    package_folder_tests::a_literal_package_folder_at_the_include_root_is_a_program_root
    package_folder_tests::a_recursive_wildcard_below_a_literal_package_folder_segment_stays_shallow
    package_folder_tests::the_other_two_package_folder_names_are_program_roots_when_named_literally
```

with e.g.

```
a_literal_package_folder_at_the_include_root_is_a_program_root
  left: []
 right: ["node_modules/top/index.ts"]
```

Closes #3395

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/3472"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

### DIFF
--- a/crates/vize/src/commands/check/tsconfig_inputs.rs
+++ b/crates/vize/src/commands/check/tsconfig_inputs.rs
@@ -23,6 +23,8 @@ mod type_references;
 #[cfg(test)]
 mod codegen_tests;
 #[cfg(test)]
+mod default_exclude_tests;
+#[cfg(test)]
 mod hidden_include_tests;
 #[cfg(test)]
 mod nuxt_manifest_tests;
@@ -47,7 +49,7 @@ use collect::{
     collect_supported_files_for_include_roots, collect_supported_files_with_options,
     explicit_hidden_include_roots, explicit_hidden_pattern_roots,
 };
-use glob::{default_exclude_specs, normalize_input_path};
+use glob::normalize_input_path;
 use loader::collect_tsconfig_project_paths;
 use matching::{
     SupportedFileOptions, is_declaration_path, is_generated_codegen_declaration_path,
@@ -144,13 +146,8 @@ fn collect_default_check_files_for_tsconfig(
         &spec.includes
     };
 
-    let default_excludes;
-    let excludes: &[GlobSpec] = if !spec.has_excludes {
-        default_excludes = default_exclude_specs(&default_base_dir);
-        &default_excludes
-    } else {
-        &spec.excludes
-    };
+    let excludes = spec.effective_excludes();
+    let excludes: &[GlobSpec] = &excludes;
 
     if !includes.is_empty() {
         let collected = collect_supported_files_for_include_roots(
--- a/crates/vize/src/commands/check/tsconfig_inputs/collect.rs
+++ b/crates/vize/src/commands/check/tsconfig_inputs/collect.rs
@@ -5,7 +5,7 @@ use std::path::{Path, PathBuf};
 use ignore::WalkBuilder;
 use vize_carton::FxHashSet;
 
-use super::glob::{default_exclude_specs, normalize_input_path, normalize_walked_path};
+use super::glob::{normalize_input_path, normalize_walked_path};
 use super::loader::{TsconfigInputCache, collect_tsconfig_project_paths};
 use super::matching::{
     SupportedFileOptions, is_generated_codegen_declaration_path, is_generated_path,
@@ -274,11 +274,7 @@ impl TsconfigOwnershipMatcher {
         } else {
             spec.includes.clone()
         };
-        let excludes = if !spec.has_excludes {
-            default_exclude_specs(&default_base_dir)
-        } else {
-            spec.excludes.clone()
-        };
+        let excludes = spec.effective_excludes();
 
         Self {
             loaded: true,
--- a/crates/vize/src/commands/check/tsconfig_inputs/default_exclude_tests.rs
+++ b/crates/vize/src/commands/check/tsconfig_inputs/default_exclude_tests.rs
@@ -1,0 +1,220 @@
+//! Tests for `tsc`'s real default `exclude` — `outDir` and `declarationDir`,
+//! and nothing else (#3395).
+//!
+//! Every expectation here was taken from `tsgo -p tsconfig.json --noEmit
+//! --listFiles` on the same tree. The three package folder names are *not* part
+//! of this default: they are rejected from wildcard `include` segments instead,
+//! which `package_folder_tests` covers.
+//!
+//! Case directories live under [`std::env::temp_dir`] for the reason spelled out
+//! in `package_folder_tests`: inside the checkout, `.gitignore` would hide
+//! `node_modules` on its own and hide the behavior under test.
+
+#![allow(clippy::disallowed_macros, clippy::disallowed_types)]
+
+use std::path::{Path, PathBuf};
+
+use vize_carton::cstr;
+
+use super::TsconfigInputCache;
+
+fn write(root: &Path, relative: &str, contents: &str) {
+    let path = root.join(relative);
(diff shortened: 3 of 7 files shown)


ubugeeei:
Compat ratchet: ran `node --test tests/tooling/compat-ratchet.test.ts` against a release build of this branch (`sha256 265e59050917e38f13d2493309d692520fa61b8dc11ca9ef3320fb2035cb2d5d`). All nine probes report `# <id> fixture is not hydrated` in this checkout (`tests/_fixtures/_git/*` is empty), so `pass 3 / skipped 9` — no baseline movement is observable locally; CI's hydrated run is the real signal.

Analytically the probes cannot move: `probeTsconfig` in `tests/_helpers/compat-ratchet.ts` writes no `exclude`, no `outDir` and no `declarationDir`, so the effective exclude set goes from the three anchored package-folder globs to empty, and every probe `include` is a nested workspace subdirectory (`template/code/typescript-default/src`, `packages/frontend/src`, …) that a root-level `node_modules/**` path cannot match. The package folders under those includes stay excluded through the wildcard-segment rule from #3401, which this PR does not touch.

Issue #3505, opened 2026-07-31
canon: `ignoreDeprecations` silences vue-tsc's TS5101 but not vize's TS5102


ubugeeei:
Found while fixing #3448, which makes vize report the option diagnostics of the user's own
`tsconfig.json`. Filed separately because the cause is the checker version skew, not the
reporting path, and the decision is a policy one.

## The divergence

`ignoreDeprecations` is the switch TypeScript gives a user to say "I have acknowledged this
deprecation". In TypeScript 6 it silences the deprecation errors; in TypeScript 7 the same
options are *removed* rather than deprecated, so it silences nothing.

Same tree, same `tsconfig.json`, under `mktemp -d`:

```json
{
  "compilerOptions": {
    "module": "ESNext",
    "moduleResolution": "bundler",
    "noEmit": true,
    "types": [],
    "baseUrl": ".",
    "ignoreDeprecations": "6.0"
  },
  "include": ["src"]
}
```

```
$ tsc --version                       # 6.0.3, the version vue-tsc 3.3.4 pins
$ tsc -p tsconfig.json --noEmit
(clean)

$ tsgo -p tsconfig.json --noEmit      # 7.0.0-dev.20260602.1, vize's checker
tsconfig.json(1,104): error TS5102: Option 'baseUrl' has been removed. Please remove it from
your configuration.
  Use '"paths": {"*": ["./*"]}' instead.
```

Without `ignoreDeprecations` the two agree that the config is an error and differ only in
the code (`TS5101` deprecated vs `TS5102` removed). With it they disagree on whether there
is a diagnostic at all.

## Why it matters now

Before #3448 vize reported no option diagnostics at all, so this was invisible. After it,
a project that did the recommended thing — acknowledge the deprecation and keep shipping on
TypeScript 6 — is clean under `vue-tsc` and reports an error under `vize check`. That is a
new false-positive class, narrower than the false negative #3448 removed, but real:
`baseUrl` plus `ignoreDeprecations` is exactly the shape a migrating project ends up in.

## Reachability

Any project on TypeScript 6 that has adopted `ignoreDeprecations` for `baseUrl`,
`target: ES5`, `moduleResolution: node`/`node10`, or `downlevelIteration`. All of those are
common in Vue projects predating the bundler-resolution era.

## Suggested direction

Decide what vize's option diagnostics are *about*. Two coherent answers:

1. **The user's toolchain.** Honor `ignoreDeprecations` for the removed/deprecated family
[...]

PR #3506, opened 2026-07-31
fix(canon): report the user's tsconfig option diagnostics

### LINKED ISSUE #3448: canon: tsconfig option diagnostics (TS5101) are never reported, so vize and vue-tsc disagree on the whole diagnostic set
Found while building vue-tsc oracles for the #3222 parity scorecard. Two related
false negatives in `tsconfig` honoring; filed together because both come from
vize not surfacing project-level (non-file) diagnostics.

## 1. Option diagnostics on the config file are never reported

`tsconfig.json` with `baseUrl` under TypeScript 6.0.3:

```json
{
  "compilerOptions": {
    "lib": ["ES2022", "DOM", "DOM.Iterable"],
    "module": "ESNext",
    "moduleResolution": "bundler",
    "noEmit": true,
    "skipLibCheck": true,
    "strict": true,
    "target": "ES2022",
    "types": [],
    "baseUrl": ".",
    "paths": { "@/*": ["./src/*"] }
  },
  "include": ["src"]
}
```

vue-tsc 3.3.4:

```
tsconfig.json(15,5): error TS5101: Option 'baseUrl' is deprecated and will stop functioning in TypeScript 7.0. Specify compilerOption '"ignoreDeprecations": "6.0"' to silence this error.
```

`vize check --format json` (build of `main` at 20408f80f): the option diagnostic
is absent.

Classification: **vize false negative**. `baseUrl` is extremely common in Vue
projects, and this is an *error*, not a warning, in TypeScript 6 — a user
migrating on vize would not learn their config breaks in TS 7 until they ran
tsc.

## 2. The consequence is worse than a missing message

Because tsc treats a config-file error as fatal for the program, vue-tsc emits
*only* TS5101 and checks nothing. vize checks the project and reports the file
diagnostics. On the same sources with the config above:

```
vue-tsc  tsconfig.json(15,5): error TS5101: ...        (1 diagnostic, no file checking)
vize     src/App.vue(6,25): error TS2345: Argument of type 'number' is not assignable to parameter of type 'string'.
```

So the two tools disagree about *which* diagnostics exist at all, in both
directions, from a single unreported config error.

The comparator already sees this coming: `collectVueTscDiagnostics` in
`tools/fixtures/typecheck-divergence.mjs` counts a project-level diagnostic into
`baselineExcludedProjectCount` and drops it, so no probe can currently notice
that vize never produces one.

## Scope note

`crates/vize_canon/src/batch/executor/cli/project_diagnostics.rs` already has a
path for project-level diagnostics (it maps `TS5023`-style config errors), so
this may be a matter of which codes are forwarded rather than new machinery.
Whether vize should *also* adopt tsc's "config error aborts the program"
behavior is the second, larger half of this issue and probably wants its own
decision: matching it is what parity means, but it converts one deprecation
notice into "no type checking at all", which is a bad default for an editor.

ubugeeei:
## The false negative

`vize check` never reported a `compilerOptions` diagnostic for an option its own
`tsconfig` sanitizer touches. On the config from #3448:

```
vue-tsc  tsconfig.json(15,5): error TS5101: Option 'baseUrl' is deprecated and will stop
                              functioning in TypeScript 7.0. ...
vize     (nothing)
```

The cause is not which codes are forwarded — `project_diagnostics::config` already forwards
config errors, and an *unknown* option is reported today. It is that
`virtual_project::tsconfig_gen` writes a **sanitized** config: `PATH_SENSITIVE_COMPILER_OPTIONS`
(`baseUrl`, `rootDir`, `rootDirs`, `tsBuildInfoFile`, …) are stripped so they cannot resolve
against the mirror, and `normalize_native_removed_options` rewrites what the native checker
removed (`target: ES5` → `ES2015`, `moduleResolution: node` → `bundler`,
`downlevelIteration` dropped). The checker never sees those options, so it never reports
them. Measured on a build of `main`:

| tsconfig | `tsgo` on the real config | `vize check` before | `vize check` after |
| --- | --- | --- | --- |
| `"baseUrl": "."` | `TS5102` | *(nothing)* | `TS5102` |
| `"target": "ES5"` | `TS5108` | *(nothing)* | `TS5108` |
| `"moduleResolution": "node"` | `TS5108` | *(nothing)* | `TS5108` |
| `"downlevelIteration": true` | `TS5102` | *(nothing)* | `TS5102` |
| `"nosuchoption": true` | `TS5023` | `TS5023` | `TS5023` (once) |

Every one of these is an **error** in TypeScript 6.0.3 too — `tsc` 6.0.3 answers `TS5107`
for `target=ES5` and `moduleResolution=node10`, and `TS5101` for `baseUrl` and
`downlevelIteration` — so this is a real divergence and not a version artifact of the
native preview. The code numbers differ because vize's checker is one major ahead.

## The fix

Ask the checker about the *unsanitized* options in a second config that builds no program:

```json
{ "compilerOptions": { …the user's options verbatim, `types` emptied… },
  "include": [], "files": [] }
```

Config options are validated before any program is built, so every option diagnostic still
comes out. Measured with `tsgo`: **45 ms** for the input-less config against 190 ms for the
same options with a single source file, and it is only written when the generated config
actually lost or rewrote something the user declared — comparing what
`tsconfig_gen` *wrote to disk* against the loaded chain, so the test can never drift from
what the checker sees. `paths` and `typeRoots` keep their key (only their targets are
re-anchored), so they never trigger it.

Only diagnostics the checker positions on that config, in TypeScript's 5xxx option range,
are forwarded; they are anchored on the project exactly like the main run's config errors,
so an option reported by both runs collapses in `dedup_diagnostics`. `TS18002` — the price
of building no program — falls outside the range and is dropped along with its
continuation lines.

`types` is the only option overridden, and only to empty, so a `@types` package that
resolves from the real tree alone cannot become a spurious `TS2688`. Nothing else is
touched: adding `noEmit` here would invent conflicts with the user's own emit options, and
a program with no inputs emits nothing anyway.

Errors found this way fail the check, as they do in `tsc`. vize still checks the project —
`tsc` treats a config error as fatal and reports nothing else — which is the deliberate
remaining difference #3448 calls out as its own decision.

`executor.rs` gets the call; the fallback-ladder machinery moves verbatim to
`executor/fallback.rs`, which takes that file from 415 to 302 lines.

## Fixtures this changed

Eight existing tests asserted a clean check on a config `tsc` rejects. Each fixture is
fixed, not re-pinned:

* `"baseUrl": "."` removed from nine fixtures — every one pairs it with `paths`, which
  resolve relative to the tsconfig directory on their own, so the fixtures are unchanged in
  meaning;
* `"moduleResolution": "Node"` → `"Bundler"` in `check_ambient_export_assignment_cli`
  (the test is about `export =` in an ambient module and never resolves from disk);
* `"declaration": true` added beside `declarationMap` in the two declaration-emit fixtures,
  which were tripping `TS5069`.

## Tests

* `crates/vize_canon/tests/tsconfig_option_diagnostics.rs` — end to end through
  `BatchTypeChecker`: a stripped `baseUrl` now reports exactly one option diagnostic on the
  user's `tsconfig.json`; a config that survives sanitization reports nothing extra *and*
  writes no probe at all; an unknown option is reported **once** beside the file
  diagnostics, which also proves the program is still checked.
* `virtual_project/option_probe/tests.rs` — the gate (dropped option, rewritten string
  option, re-anchored `paths`, unchanged options) and the exact probe config.
* `executor/option_probe/tests.rs` — parsing verbatim `tsgo` output: every 5xxx code with
  its continuation line, a dropped `TS18002` taking its continuation with it, diagnostics on
  any other file ignored, and the absolute-path spelling of the config.

**Failing before the fix**: with the probe call in `check_with_servers` replaced by an empty
diagnostic list, `a_stripped_path_sensitive_option_still_reports_its_diagnostic` fails with
`expected one option diagnostic: []` / `left: 0, right: 1`. The other two end-to-end cases
assert an absence and a de-duplication, so they hold trivially without the probe and are
there to pin that it adds nothing it should not — as the eight fixture tests above, which
passed only because vize was silent, now demonstrate from the other side.

## Not fixed here

`"ignoreDeprecations": "6.0"` silences `TS5101`/`TS5107` in TypeScript 6 but does not
silence the native preview's `TS5102`/`TS5108`, because in TypeScript 7 those options are
gone rather than deprecated. A project that has correctly acknowledged the deprecation is
therefore clean under `vue-tsc` and reports an error under vize. Filed separately as
#3505.

Closes #3448

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/3506"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

### DIFF
--- a/crates/vize/tests/check_ambient_export_assignment_cli.rs
+++ b/crates/vize/tests/check_ambient_export_assignment_cli.rs
@@ -57,7 +57,7 @@ fn check_allows_export_assignment_inside_ambient_module_declaration() {
     "strict": true,
     "target": "ES2022",
     "module": "ESNext",
-    "moduleResolution": "Node",
+    "moduleResolution": "Bundler",
     "noEmit": true
   },
   "include": ["src/**/*.d.ts"]
--- a/crates/vize/tests/check_canon_graphql_cli.rs
+++ b/crates/vize/tests/check_canon_graphql_cli.rs
@@ -137,7 +137,6 @@ fn check_explicit_vue_keeps_generated_graphql_schema_out_of_canon() {
     "target": "ES2022",
     "module": "ESNext",
     "moduleResolution": "bundler",
-    "baseUrl": ".",
     "paths": {
       "~/*": ["*"]
     },
@@ -243,7 +242,6 @@ fn check_barrel_reexport_preserves_generated_graphql_identity() {
     "target": "ES2022",
     "module": "ESNext",
     "moduleResolution": "bundler",
-    "baseUrl": ".",
     "paths": {
       "~/*": ["*"],
       "@/*": ["*"]
--- a/crates/vize/tests/check_cli.rs
+++ b/crates/vize/tests/check_cli.rs
@@ -1183,7 +1183,6 @@ const configs: AliasConfig[] = [{ key: "a", label: "A" }];
     "module": "ESNext",
     "moduleResolution": "bundler",
     "noEmit": true,
-    "baseUrl": ".",
     "paths": { "@/*": ["src/*"] }
   },
   "include": ["src/**/*"]
@@ -1483,7 +1482,6 @@ const same: Registry["home"] = entry;
     "module": "ESNext",
     "moduleResolution": "bundler",
     "noEmit": true,
-    "baseUrl": ".",
     "paths": {
       "#registry": ["./registry"]
     }
@@ -3106,6 +3104,7 @@ fn check_declaration_emit_uses_tsconfig_options() {
     "module": "ESNext",
     "moduleResolution": "bundler",
     "noEmit": true,
+    "declaration": true,
     "declarationDir": "types-from-tsconfig",
     "declarationMap": true
   },
@@ -3211,7 +3210,6 @@ fn check_json_handles_monorepo_tsconfig_extends_paths_and_excludes() {
     "target": "ES2022",
     "module": "ESNext",
     "moduleResolution": "bundler",
-    "baseUrl": ".",
     "paths": {
       "@shared/*": ["packages/shared/src/*"]
     },
--- a/crates/vize/tests/check_declaration_emit_cli.rs
+++ b/crates/vize/tests/check_declaration_emit_cli.rs
@@ -240,6 +240,7 @@ defineProps<PublicProps>()
     "target": "ES2022",
     "module": "ESNext",
     "moduleResolution": "bundler",
+    "declaration": true,
     "declarationMap": true,
     "sourceMap": true,
     "inlineSources": true,
--- a/crates/vize/tests/check_nuxt2_plugin_injections_cli.rs
+++ b/crates/vize/tests/check_nuxt2_plugin_injections_cli.rs
@@ -20,7 +20,6 @@ fn check_nuxt2_use_context_sees_plugin_injections() {
     "target": "ES2022",
     "module": "ESNext",
     "moduleResolution": "bundler",
-    "baseUrl": ".",
     "noEmit": true
   },
   "include": ["pages/**/*.vue", "plugins/**/*.ts", "types/**/*.d.ts"]
--- a/crates/vize/tests/check_nuxt_tsconfig_paths_cli.rs
+++ b/crates/vize/tests/check_nuxt_tsconfig_paths_cli.rs
@@ -20,7 +20,6 @@ fn check_nuxt_sfc_virtual_ts_prefers_explicit_tsconfig_paths_over_fallback_modul
     "target": "ES2022",
     "module": "ESNext",
     "moduleResolution": "bundler",
-    "baseUrl": ".",
     "noEmit": true,
     "paths": {
       "#imports": ["types/vize/imports.ts"],
@@ -178,7 +177,6 @@ fn check_nuxt2_options_api_component_event_payloads_resolve_through_aliases() {
     "target": "ES2022",
     "module": "ESNext",
     "moduleResolution": "bundler",
-    "baseUrl": ".",
     "noEmit": true,
     "noUnusedLocals": true,
     "paths": {
--- a/crates/vize/tests/check_package_cwd_cli.rs
+++ b/crates/vize/tests/check_package_cwd_cli.rs
@@ -76,7 +76,6 @@ fn check_from_package_cwd_uses_package_local_tsconfig_inputs() {
     "target": "ES2022",
     "module": "ESNext",
     "moduleResolution": "bundler",
-    "baseUrl": ".",
     "paths": {
       "~/*": ["*"]
     },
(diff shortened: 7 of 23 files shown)


ubugeeei:
Converted to **draft** and auto-merge disabled: `vue-parity` is red for a reason this PR cannot resolve on its own.

Seven pinned real-world fixtures now report `TS5090: Non-relative paths are not allowed. Did you forget a leading './'?` alongside the intended `TS5102`. That code exists only because TypeScript 7 removed `baseUrl`, which makes a `paths` target without a leading `./` invalid — TypeScript 6.0.3, the version `vue-tsc` 3.3.4 pins, accepts the same config. `baseUrl` plus non-relative `paths` is the most common `paths` spelling in Vue projects, so forwarding the checker's verdict wholesale trades the false negative in #3448 for a false positive against the tool the parity scorecard measures against. `"ignoreDeprecations": "6.0"` has the same shape (#3505).

Full analysis, the measured blast radius, and the two ways forward are on #3448: https://github.com/ubugeeei-prod/vize/issues/3448#issuecomment-5138237422

The mechanism itself works and is fully tested — the probe config, the gate, the 5xxx parsing and the three end-to-end cases. What needs a decision before it can land is whether a vize option diagnostic speaks for the user's `vue-tsc` toolchain or for the TypeScript 7 checker vize actually runs. The fixture fixes in this PR (eight tests asserting a clean check on configs `tsc` also rejects) are uncontroversial and could be split out first.

ubugeeei:
Superseded by #3524, which takes this implementation unchanged and adds the narrowing that settles the open question: a vize option diagnostic represents the user's `vue-tsc` toolchain, not vize's own checker (option 1 from the analysis above).

Also corrects the fixture half of this PR — dropping `baseUrl` is only sound when the `paths` target already resolves relative to the tsconfig directory, and `"src/*"` does not. Verified against tsgo: it is `TS5090` with no `baseUrl` under *both* compilers, so the four fixtures were left carrying configs `tsc` and `vue-tsc` both reject. `"./src/*"` is clean.

`vue-parity` is clean with no re-pins (measured against the hydrated fixtures).

ubugeeei:
Superseded by #3524, which landed this implementation together with the narrowing that settles the open question — a vize option diagnostic represents the user's `vue-tsc` toolchain, not vize's own checker.

It also corrects the fixture half: dropping `baseUrl` is only sound when the `paths` target already resolves relative to the tsconfig directory, and `"src/*"` does not. Verified against tsgo — it is `TS5090` with no `baseUrl` under *both* compilers, so the four fixtures here were left carrying configs `tsc` and `vue-tsc` both reject. `"./src/*"` is clean.

`vue-parity` is clean with no re-pins, measured against the hydrated fixtures.

Issue #3460, opened 2026-07-30
perf: lint and fmt medians regressed ~20-27% between 27fe77df and 1511788d9


ubugeeei:
Surfaced by #3459, which refreshed the Blacksmith snapshot from
[run 30557718030](https://github.com/ubugeeei-prod/vize/actions/runs/30557718030) at `1511788d9`.
The previous artifact was [run 29010164013](https://github.com/ubugeeei-prod/vize/actions/runs/29010164013)
at `27fe77df` (2026-07-09).

Same runner label (`blacksmith-32vcpu-ubuntu-2404`), same corpus size (15,000 SFCs), same surfaces.

## The movement

| Surface | Vize before | Vize after | change | baseline before | baseline after |
| --- | ---: | ---: | ---: | ---: | ---: |
| Lint | 270.9ms | 324.8ms | **+20%** | 56.66s | 56.20s |
| Format | 2.22s | 2.83s | **+27%** | 150.34s | 143.13s |
| SFC compile | 333.8ms | 329.2ms | flat | 18.38s | 17.15s |

The baselines are near-identical across the two runs, and SFC compile is flat, so this is not a
uniform runner slowdown — it is isolated to the `vize lint` and `vize fmt` maxima.

## Why it is probably not the benchmark-fairness work

The two harness fixes in this window (#3392, #3409) changed only the Vite and Nuxt lanes:
#3392 clears the pre-compile manifest for `bench/vite.ts`, and #3409 clears `.vize`/`.cache`/`.vite`
under the shared Nuxt `node_modules`. Neither touches the lint or fmt measurement paths, which run
through `bench/compare-tools.mjs`'s own per-invocation work directories.

## Caveat, stated plainly

Two Blacksmith runs three weeks apart are **not** proof of a regression. Runner variance, a changed
base image, or a dependency bump could account for some of it. What makes it worth chasing is that
the two moved surfaces moved a lot, in the same direction, while their baselines and a third Vize
surface did not.

## Suggested next step

Bisect with the criterion lanes rather than the end-to-end tool comparison, since those are cheaper
and already have baselines: `bench/criterion-baselines.mjs`, `bench/criterion-ab.mjs`, and the
`lint_bench.rs` / `formatter.rs` criterion groups. If the regression reproduces there, the commit
range `27fe77df..1511788d9` is the search space.

Note also that `bench/results/tool-benchmark-latest.json` now records binary sha256 (#3434), so a
[...]

PR #3510, opened 2026-07-31
perf(canon): hoist the template ref-widening helpers out of every module

ubugeeei:
## The defect

`__VizeIsUnion` and `__VizeWidenTemplateRef` take every input as a type
parameter, so they are byte-identical in every generated virtual module. They
were nevertheless declared **per file**, inside `__template()`:

```ts
  ;(function __template() {
    // Auto-unwrap Vue refs in template scope
    type __VizeIsUnion<T, __U = T> = …;              //  ~101 B  ← file-independent
    type __VizeWidenTemplateRef<T> = …;              //   268 B  ← file-independent
    type __U<T> = T extends import('vue').Ref ? __VizeWidenTemplateRef<T['value']> : T;
```

That is the same shape #3443 removed for `__VizeNativeElement` /
`__VizeNativeElementProp`, where hoisting the per-file conditional types into
the program-wide `__vize_helpers.d.ts` was worth ~50%: a per-file declaration is
a distinct declaration as far as TypeScript's instantiation cache is concerned,
so nothing is shared across the program.

It was found while bisecting #3460. `__VizeWidenTemplateRef` is the single
largest per-file helper in the generated preamble — #2957 grew it from 162 to
268 bytes, which is 27% of the +6.2% virtual-TypeScript growth measured across
`accc7194f..1511788d9`.

## The change

`SHARED_PREAMBLE_DTS` declares both aliases itself, so the batch path emits
them **once per program** in the hoisted ambient helpers file and each generated
module keeps only its own `__U`.

`__U` **stays per file**: it is dialect-dependent — Vue 2, Vue 3, and Vue 3 with
`generic` each need a different body (`LEGACY_REF_UNWRAP_HELPER`,
`MODERN_REF_UNWRAP_HELPER`, `MODERN_GENERIC_REF_UNWRAP_HELPER`) — so each module
keeps its own and delegates to the hoisted conditional types.

### The non-hoisted path is left byte-identical to `main`

The first push declared the aliases in `vue_type_aliases_text!()` instead, which
also put them at **module scope** on the path that does not hoist (check server,
content mapper). `clippy-and-test` caught the regression that follows, and this
is the substance of the second commit:

```
---- check_server_maps_dependency_patch_diagnostics_to_the_parent_sfc ----
  left: [{ "code": "TS6196", "line": 39, "column": 159,
           "message": "'__VizeWidenTemplateRef' is declared but never used.",
           "severity": "hint" }]
 right: []
```

Module-scope declarations are module-local, so an unused one is reported;
declarations in the hoisted `.d.ts` global script are ambient and are not.
`__VizeWidenTemplateRef` is the only alias in that block with no unconditional
consumer: `__VizeIsAny` is used by `__VForEntry` and `__VizeIsUnion` by
`__VizeWidenTemplateRef`, but `__VizeWidenTemplateRef` is used only by `__U`,
and `emit_template_variables` returns before emitting `__U` for a component with
no setup binding in template scope. That component would carry a dead
module-scope alias, and it surfaces to the user as a hint on their own file.

So the two emissions are separate rather than sharing one constant:
`MODERN_REF_UNWRAP_HELPER` keeps all three declarations inside `__template()`,
`MODERN_HOISTED_REF_UNWRAP_HELPER` is the `__U` line alone, and
`hoist_shared_preamble` is threaded through `ref_unwrap_helper_for_template` to
choose between them. Keeping the aliases out of `vue_type_aliases_text!()` also
keeps them out of `DECLARATION_HELPERS_DTS`, where emitted `.d.ts` files never
referenced them.

The 37 virtual-TS snapshots the first commit had to update are therefore back to
`main`'s bytes; the only two snapshots this PR still changes are the hoisted
ones (`pruned_helpers`, `ref_arity__dialect_default_v3`).

## Measurement

Re-run for this push. 5 SFCs from `bench/generate.mjs`
(`generateCorpus({ fileCount: 5 })`) in one **fixed** directory for both arms,
two pinned `--profile ci` binaries built from a private `CARGO_TARGET_DIR`:

```
before  9bc3b16cb87a36b84ba86d3fccca557515fc27982c84a563f1b84ec148aa0f24
after   60c8f5d564114be77e80bf287044ce8cc495f7c1a11b831801de3b5490a6b91b

$BIN check . --quiet --tsconfig tsconfig.json --corsa-path <tsgo> \
  --save-virtual-ts-for Component000N.vue --save-virtual-ts-for __vize_helpers.d.ts
```

| file | before | after | delta |
| --- | ---: | ---: | ---: |
| Component0000 | 3,451 | 3,082 | −369 |
| Component0001 | 3,634 | 3,265 | −369 |
| Component0002 | 5,263 | 4,894 | −369 |
| Component0003 | 10,182 | 9,813 | −369 |
| Component0004 | 12,659 | 12,290 | −369 |
| **total (modules)** | **35,189** | **33,344** | **−1,845 (−5.2%)** |
| `__vize_helpers.d.ts` | 8,220 | 8,815 | +595 |

A flat −369 bytes per generated module, which is what a per-file constant
should look like when it stops being per-file, against +595 bytes **once per
program** — 361 for the two alias lines and 233 for the comment recording why
they live only there. Net negative from the second module onward; the corpus
`bench/compare-tools.mjs` type-checks is 15,000 SFCs.

Output equivalence was verified rather than asserted: `vize check . --format
json` over the same corpus is byte-identical between the two binaries
(`sha256 0009a38c…` both ways), and both report `No type errors found!`.

## The tests that fail before the fix

- `template_ref_widening_helpers_are_hoisted_once_per_program`
  (`batch/virtual_project/tests/ref_arity.rs`) registers a three-file Vue 3
  project and asserts an exact count, not a substring: each generated module
  declares the hoisted aliases **0** times and its own `__U` **1** time, and
  `SHARED_PREAMBLE_DTS` declares each of them exactly **1** time. Against
  `main` it fails with `left: 1, right: 0` on `type __VizeIsUnion<`.
- `widening_helpers_are_declared_with_the_template_scope_that_uses_them` and
  `a_component_without_setup_bindings_declares_no_widening_helpers`
  (`batch/virtual_project/content_mapper_tests.rs`) pin the non-hoisted half —
  the content mapper is that path. The first asserts the **full** four-line
  `__template()` prelude with `assert_eq!`; the second asserts that a component
  with nothing to unwrap declares no `__U` and neither alias. Verified against
  the first commit by reverting only the four source files
  (`git checkout HEAD~1 -- …`) and re-running:

  ```
  ---- a_component_without_setup_bindings_declares_no_widening_helpers ----
  assertion `left == right` failed: type __VizeIsUnion< would be an unused
    module-local declaration
    left: 1
   right: 0

  ---- widening_helpers_are_declared_with_the_template_scope_that_uses_them ----
    left: "    // Auto-unwrap Vue refs in template scope\n    type __U<T> = …"
   right: "    // Auto-unwrap Vue refs in template scope\n    type __VizeIsUnion<T, …"
  ```
- `crates/vize/tests/check_ref_enum_template_cli.rs` carries the CLI-level half
  of the invariant: the saved virtual TS still unwraps through
  `__VizeWidenTemplateRef` and re-declares it **0** times.

## Gates

Run in a private `CARGO_TARGET_DIR` with `CARGO_INCREMENTAL=0`, on the rebased
tree:

- `cargo test -p vize_canon` → `727 passed; 0 failed`, no snapshot drift.
- `cargo test -p vize --test check_server_cli --test check_cli --test
  build_declaration_emit_cli` → `1 + 45 + 4 passed; 0 failed`. The check-server
  test is the one that failed on the first push; it runs for real here
  (`CORSA_PATH` pointed at a `tsgo`), it does not skip.
- `cargo clippy --workspace -- -D warnings -D clippy::wildcard_imports` → clean.
- `cargo check --workspace` → clean.
- `rustfmt 1.9.0 --edition 2024 --config style_edition=2024 --check` on all five
  changed source files → clean.
- Source-length gate → `No new or grown files exceed 350 lines.` `helpers.rs`
  stays at exactly 598 lines (over budget, so it may not grow) and
  `generator.rs` drops 973 → 972: the two-line comment at the
  `emit_template_variables` call site moved onto the method it describes, which
  is also where the reason for the split now lives.
- `tests/snapshots/check/*.snap` are unchanged by the second commit. They were
  regenerated for the first one, and the hoisted per-module output is identical
  between the two — `ref_arity__dialect_default_v3.snap` (hoisted) is unchanged
  by the second commit as well, which is the proof.

Refs #3460

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved template reference type handling across generated Vue projects.
  - Prevented duplicate shared helper declarations across modules.
  - Ensured ref-unwrapping helpers are emitted only when required and in the correct template scope.

- **Tests**
  - Added regression coverage for template-reference widening, helper hoisting, module deduplication, and components without setup bindings.


ubugeeei:
One consequence worth flagging for review: the two aliases live in
`vue_type_aliases_text!()`, which also feeds the **declaration-emit** shared
helpers file. `.d.ts` output has no template scope and never references `__U`,
so those two lines are unused there — two lines, once per program, in exchange
for a single source of truth for the alias text. Splitting them into
`vue_type_helpers_text!()` would avoid it, but `helpers.rs` is already over the
350-line budget and may not grow, and the existing packed-line style is what
keeps this change at zero net lines there.

Also note `tests/snapshots/check/__snapshots__/ecosystem-products-check.snap`
was derived mechanically rather than by re-running the fixture: that fixture is
environment-sensitive here (two consecutive local runs disagreed on a
`TS2307 Cannot find module '@formkit/vue'`), so regenerating it locally would
have baked in a resolution artifact. The committed version is `main`'s file with
exactly the two alias lines deleted from each embedded `virtualTs` — 5
occurrences, 1,855 bytes, no other edit. The other three check snapshots were
regenerated by running the fixtures.

ubugeeei:
## Moved to draft: this breaks the **non-hoisted** path, and the fix is a bigger change than the PR currently is

`clippy-and-test` caught a real regression, not a snapshot lag:

```
---- check_server_maps_dependency_patch_diagnostics_to_the_parent_sfc ----
  left: [{ "code": "TS6196", "line": 39, "column": 159,
           "message": "'__VizeWidenTemplateRef' is declared but never used.",
           "severity": "hint" }]
 right: []
```

(`crates/vize/tests/check_server_cli.rs:118`.)

### Why

The change is correct for the hoisted path and wrong for the other one.

- **Hoisted** (`vize check` batch): the aliases land in `__vize_helpers.d.ts`,
  a global script. Unused declarations there are not reported. This is the path
  the −369 B/file measurement came from, and it is fine.
- **Non-hoisted** (check server, content-mapper, Vue 2 dialect): `VUE_TYPE_HELPERS`
  is embedded at **module** scope of each generated `.vue.ts`. Module-scope
  declarations are module-local, so an unused one is reported.

`__VizeWidenTemplateRef` is the only alias in that block with no unconditional
consumer: `__VizeIsAny` is used by `__VForEntry` and `__VizeIsUnion` by
`__VizeWidenTemplateRef`, but `__VizeWidenTemplateRef` is used only by `__U`,
and `__U` is emitted only when the component actually has setup bindings in
template scope (`emit_template_variables` returns early otherwise). A component
without them now carries a dead module-scope alias — and it surfaces to the user
as a hint diagnostic.

### What the fix has to be

Split the two emissions rather than sharing one constant:

1. Drop the aliases from `vue_type_aliases_text!()` (so the non-hoisted
   module-scope preamble does not carry them).
2. Add them to `SHARED_PREAMBLE_DTS` only — appended to the existing
   `VUE_TYPE_HELPERS,` concat entry, so `helpers.rs` stays at 598 lines.
3. Keep the current 3-line `MODERN_REF_UNWRAP_HELPER` for the non-hoisted path
   and add a hoisted variant that is only the `__U` line, selected by threading
   `hoist_shared_preamble` into `ref_unwrap_helper_for_template`
   (`generator/legacy_vue2.rs`, called from `generator/template_refs.rs:92`).

That keeps the whole win — the hoisted batch path is the one that generates
thousands of modules — and leaves the non-hoisted path byte-identical to today.
It also means another full snapshot regeneration cycle.

I am out of budget for this pass, so: auto-merge disabled, moved to draft, and
the branch left as the measurement plus the failing case. The measurement itself
stands and is reproducible — `main` + this change is 33,344 B against 35,189 B
on the same 5-SFC corpus, a flat −369 per generated module.

Refs #3460

ubugeeei:
The fix from the previous comment is pushed (`cbfd3c3f0`), with `origin/main` merged in so the push could stay fast-forward.

`SHARED_PREAMBLE_DTS` now declares `__VizeIsUnion` / `__VizeWidenTemplateRef` itself instead of inheriting them from `vue_type_aliases_text!()`, `MODERN_HOISTED_REF_UNWRAP_HELPER` is the `__U` line alone, and `hoist_shared_preamble` is threaded through `ref_unwrap_helper_for_template`. The non-hoisted path is byte-identical to `main` again — the 37 virtual-TS snapshots the first commit updated are reverted, and `check_server_maps_dependency_patch_diagnostics_to_the_parent_sfc` passes locally against a real `tsgo` (it does not skip).

Two additions since the write-up:

- The two aliases are also out of `DECLARATION_HELPERS_DTS` now, which retires the wart flagged earlier in this thread: emitted `.d.ts` files never referenced them, and they were costing two dead lines per declaration-output directory.
- The measurement is re-run rather than carried over, and the PR body now also reports the cost side: `__vize_helpers.d.ts` grows 8,220 → 8,815 B once per program against −369 B per module, so the change is net negative from the second module onward. `vize check --format json` over the corpus is byte-identical between the two pinned binaries.

Left in **draft with auto-merge disabled** for the release freeze, not because anything is outstanding. Ready to un-draft and arm `--auto --squash` as soon as the freeze lifts.

ubugeeei:
CI is fully green on `cbfd3c3f0`: **26 checks pass, 0 fail**, including `clippy-and-test` (the job that failed on the first push), `test-scripts`, `tool-benchmark`, `pr-benchmark` and `vue-parity`.

Staying in draft only because of the release freeze — a draft cannot be merged, which is the strongest guarantee I can give while a tag is in flight. No auto-merge is armed. Un-draft and `gh pr merge 3510 --auto --squash` is all that is left.

Issue #3446, opened 2026-07-30
canon: generic component props emit a spurious TS7006 and anchor two TS2322 at the wrong bytes


ubugeeei:
Found while building vue-tsc oracles for the #3222 parity scorecard. This is the
worst single case in the battery: three vize-only diagnostics and two missed
ones on a five-line template, mixing a genuine false positive with two span
divergences.

## Reproduction

`src/Child.vue`

```vue
<script setup lang="ts" generic="T extends { id: number }">
defineProps<{ items: T[]; pick: (item: T) => string }>()
</script>

<template><span /></template>
```

`src/Parent.vue`

```vue
<script setup lang="ts">
import Child from './Child.vue'
</script>

<template>
  <Child :items="[{ id: 1 }]" :pick="(item) => item.id" />
  <Child :items="[{ id: 'x' }]" :pick="() => 'y'" />
</template>
```

`tsconfig.json`: `strict: true`, `moduleResolution: "bundler"`, `noEmit: true`,
`lib: ["ES2022", "DOM", "DOM.Iterable"]`, `types: []`.

## Oracle — vue-tsc 3.3.4

```
src/Parent.vue(6,48): error TS2322: Type 'number' is not assignable to type 'string'.
src/Parent.vue(7,21): error TS2322: Type 'string' is not assignable to type 'number'.
```

## vize check (build of `main` at 20408f80f)

```
src/Parent.vue(6,31): error TS2322: Type '(item: { id: number; }) => number' is not assignable to type '(item: { id: number; }) => string'. Type 'number' is not assignable to type 'string'.
src/Parent.vue(6,39): error TS7006: Parameter 'item' implicitly has an 'any' type.
src/Parent.vue(7,22): error TS2322: Type 'string' is not assignable to type 'number'.
```

## Classification

1. **Vize false positive — `6,39` `TS7006: Parameter 'item' implicitly has an
   'any' type`.** vue-tsc infers `item` from the child's generic parameter, so
   the inline arrow's parameter is contextually typed and there is no implicit
   `any`. This is the highest-value part: it is a *new error on correct code*
   under `strict`, and inline callbacks in generic-component props are common.

2. **Span divergence — line 6.** Both tools flag the same mistake, at different
   places: vue-tsc anchors at column 48 (inside the arrow body, at `item.id`,
   the expression whose type is wrong), vize at column 31 (the `pick` attribute
   name) and with a whole-signature message instead of the leaf mismatch.

[...]

PR #3511, opened 2026-07-31
fix(canon): contextually type an inline callback prop on a generic child

ubugeeei:
Closes the false positive and the off-by-one from #3446. The third item — the
line-6 anchor — is analysed below and left open.

## Reproduction

`src/Child.vue` is `generic="T extends { id: number }"` with
`defineProps<{ items: T[]; pick: (item: T) => string }>()`; `src/Parent.vue`
lines 6-7 are

```vue
  <Child :items="[{ id: 1 }]" :pick="(item) => item.id" />
  <Child :items="[{ id: 'x' }]" :pick="() => 'y'" />
```

vue-tsc 3.3.4:

```
src/Parent.vue(6,48): error TS2322: Type 'number' is not assignable to type 'string'.
src/Parent.vue(7,21): error TS2322: Type 'string' is not assignable to type 'number'.
```

vize before:

```
src/Parent.vue(6,31): error TS2322: Type '(item: { id: number; }) => number' is not assignable to type '(item: { id: number; }) => string'. …
src/Parent.vue(6,39): error TS7006: Parameter 'item' implicitly has an 'any' type.
src/Parent.vue(7,22): error TS2322: Type 'string' is not assignable to type 'number'.
```

## 1. The `TS7006` false positive

A generic child is checked through its `__vizeCheck<T>(props)` call, so
`__X_Props_N` resolves to `Record<string, unknown>` and every per-prop alias to
`unknown`. The per-prop check therefore generates

```ts
const __vize_prop_check_0_pick: unknown = (item) => item.id;
```

An `unknown` annotation gives the arrow no contextual type at all, so `strict`
reports `TS7006` on a parameter vue-tsc infers from the child's generic — a new
error on correct code, on a shape that is common. (I confirmed against tsgo that
neither `unknown` nor `any` as the annotation suppresses it; only a callable
contextual type does.)

The per-prop alias now falls back to a permissive callable **when the authored
value is an inline function**:

```ts
type __VizeCallableProp<T> = __VizeIsAny<T> extends true ? T : unknown extends T ? (...args: any[]) => any : T;
```

- `any` is excluded, so a genuinely `any` prop stays assignable from a
  non-function value (without the guard, `CallableProp<any>` rejects
  `[{ id: 1 }]`).
- A resolved prop type is returned untouched, so a real mismatch on a
  *non*-generic child still surfaces exactly where it did.
- Gating on the authored value being an inline function keeps every other prop's
  annotation unchanged, so a non-function value can never be checked against a
  function type.

## 2. The off-by-one on line 7

The generic checker call mapped the whole `"prop": value` entry onto the whole
attribute with no sub-spans, and `"prop": ` is one byte longer than `:prop="`,
so an error *inside* the value (TypeScript anchors a nested object-literal
mismatch at the offending key) landed one byte to the right. Mapping the key and
the value separately makes the value range verbatim, so offsets inside it are
exact: `7:22` -> `7:21`, byte-identical with vue-tsc.

## Not fixed

The line-6 assignability error is still reported as a whole-signature mismatch
at the `pick` attribute name (now column 32, the attribute name, rather than
column 31, the `:`) where vue-tsc reports the leaf inside the arrow body at
column 48. vue-tsc gets there because it resolves the child's instantiated prop
type per prop, so the arrow body is checked in place; vize only resolves the
generic inside the `__vizeCheck` call, where TypeScript anchors object-literal
property errors at the key. Closing it means giving the per-prop check the
resolved type — plausibly through the `__vizeResolveProps` helper the child
already exports — which is a separate change with its own inference-order risk.
Recorded on the issue; this PR is `Refs #3446`, not `Closes`.

## Verification

New `crates/vize_canon/src/batch/type_checker/tests/generic_props/inline_callback_props.rs`
asserts the **complete** diagnostic list, with the vue-tsc oracle in the doc
comment. On the pre-fix code (`git checkout HEAD~1 -- <the three changed source
files>`):

```
  left:  [(… 2322, "6:31:error …"), (… 2322, "7:22:error …"), (… 7006, "6:39:error Parameter 'item' implicitly has an 'any' type.")]
  right: [(… 2322, "6:32:error …"), (… 2322, "7:21:error …")]
```

Gates run locally (macOS, `nix develop -c`):

- `cargo test -p vize_canon --lib` — 662 passed, 0 failed
- `cargo clippy -p vize_canon -- -D warnings -D clippy::wildcard_imports` — clean
- `rustfmt --edition 2024 --config style_edition=2024 --check` on every changed `.rs` — clean
- `node --test tests/tooling/source-file-lengths.test.ts` — 7/7 pass, no over-limit file grew
- `node --test tests/tooling/compat-ratchet.test.ts` with all nine `vue-parity`
  fixtures hydrated — 12/12 pass, every probe still at
  `messageMismatches=0 falsePositives=0 falseNegatives=0`
- `tests/snapshots/check/{typecheck-errors,compiler-macros,ecosystem-products,generic-build,options-api}.ts` — all pass

The binary used for the CLI suites was built in a **private** target dir
(`cargo build --profile ci -p vize`, sha256 recorded), after the shared one
turned out to have been overwritten by a concurrent build.

Every generated-output snapshot delta is the one added `__VizeCallableProp` line;
no diagnostic or `errorCount` moved in any of them.

Refs #3446

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved type checking for inline callback props on generic components.
  * Diagnostics now point more precisely to relevant prop names and values.
  * Prevented spurious implicit-`any` errors while preserving valid type errors.
  * Improved checking and source mapping for generic component props, including conditional and spread bindings.
  * Improved handling of class bindings and spread properties in generic prop diagnostics.

* **Tests**
  * Added regression coverage for contextual typing and diagnostics involving generic component callback props.


ubugeeei:
Rebased onto main and resolved against #3525 (`__VizeExactOptionalProps`), which had landed in the same helper function and the same per-usage alias.

- `append_prop_helper_types` and `append_prop_check_helpers` were the same function under two names; kept the latter and grafted `__VizeCallableProp` onto it, emitted last.
- `prop_alias_type` and the per-prop alias loop now live together in `component_prop_checker.rs`.
- Two source-length breaches the rebase created are fixed by extraction rather than by shrinking the change: `expressions/component_props.rs` crossed 350 (312 -> 360), so the generic prop-checker call moved to `expressions/generic_props_call.rs`; `scope/component_props.rs` was already over budget and grew 397 -> 399, so the alias loop moving out takes it to 379.
- `ecosystem-products-check.snap` is rebuilt on #3530's CI-verified post-#3525 content with `__VizeCallableProp` inserted in emission order. That fixture does not reproduce locally — it needs check options that differ outside CI — so this is the only sound way to pin it.

23 `vize_canon` test binaries, clippy and the source-length gate are green; `vue2-elm` and the `vue-element-admin` oracles pass unchanged.

ubugeeei:
Fixed a real defect CI caught: `__VizeCallableProp` was emitted unconditionally but only referenced when a prop is an inline callback, so a program without one leaked

```
TS6196: '__VizeCallableProp' is declared but never used.
```

to `check-server` clients as an unmapped hint on a clean SFC — `check_server_maps_dependency_patch_diagnostics_to_the_parent_sfc`. Same failure mode the native element aliases had before #3443.

The ambient `declare function` those use is not available here: these helpers are emitted inside a template scope's function body rather than at module level. So the alias is now emitted only when some usage actually binds an inline callback.

The emission gate and the alias loop share one predicate (`is_inline_callback_prop`, which carries the `key`/`ref`/dynamic-name filters), because the two must agree — emitting the helper for a prop the loop then skips reproduces exactly this bug. That was the first fix attempt and it still failed, which is what made the shared predicate necessary rather than merely tidy.

Green: 23 `vize_canon` test binaries, `check_server_cli`, `check_cli` (45), the declaration-emit vue-floor consumer, clippy and the source-length gate.

Issue #3516, opened 2026-07-31
ci: app check/lint snapshots are only validated at release time


ubugeeei:
## The shape of the problem

`tests/snapshots/check/__snapshots__/{elk,misskey,npmx.dev,nuxt-ui,reka-ui}-check.snap`
and `tests/snapshots/lint/__snapshots__/*.snap` are asserted **only** by the
`App E2E` `check` / `lint` suites, and those run only on `schedule`,
`workflow_dispatch`, and the tag/release dispatch. Nothing validates them on a PR.

So a PR that legitimately changes diagnostics is green, merges, and the drift is
invisible until someone cuts a release — at which point `release-preflight` is
already running and the tag is burned.

That is exactly what happened for v0.307.0 and v0.308.0. Five PRs merged between
the last green run (v0.306.0, `8c2509a23`, run 30567979170) and v0.308.0
(`83343426f`, run 30602594088), and all five of them intentionally change
diagnostics:

| PR | what it changes |
| --- | --- |
| #3443 | native element `v-bind` is now type-checked for every attribute name |
| #3483 | module diagnostics quote the authored `.vue`, not the `.vue.ts` mirror |
| #3509 | handler / `v-model` diagnostics anchor at the directive name (`TS2345` → `TS2322`) |
| #3510 | ref-widening helpers hoisted program-wide (shifts unanchored positions) |
| #3512 | `@click="functionalHandler(evt)"` is no longer a function expression |

Each of them updated the *fixture* check snapshots it broke — those are gated on
PRs by `check.yml`'s `vue-parity` job (`test:check:fixtures`) — but none of them
could have known about the five app snapshots. #3508 did the same to the `lint`
snapshots.

Re-pinned in #3515 (check). The `lint` side is still red.

## Concrete cheap proposal

`e2e.yml`'s `app-readiness` job already runs on every PR, already hydrates the
`elk` and `misskey` submodules, and already builds the `vize` CLI. `elk`'s check
needs no fixture install at all — its `check.cwd` is the pristine submodule — and
takes **~730 ms** (measured in run 30602594088). So the smallest useful version
is one line in `tests/package.json`:

```diff
-    "test:readiness:check": "node --test --test-concurrency=1 snapshots/check/compiler-macros.ts",
+    "test:readiness:check": "node --test --test-concurrency=1 snapshots/check/compiler-macros.ts snapshots/check/elk.ts",
```

plus the two matching entries in the `readiness` paths-filter in `e2e.yml`:

```
[...]

PR #3515, opened 2026-07-31
test(canon): re-pin the five app check snapshots to the landed diagnostic fixes

ubugeeei:
## The failure

`App E2E (check)` failed on the v0.308.0 release commit `83343426f`
([run 30602594088](https://github.com/ubugeeei-prod/vize/actions/runs/30602594088)):
elk, misskey, npmx.dev, nuxt-ui and reka-ui all failed
`vize check does not crash and snapshot matches`.

That suite only runs on tag / nightly / dispatch. It last ran green at v0.306.0
(`8c2509a23`), and five PRs that each intentionally change diagnostics have
merged since, so the drift accumulated invisibly and landed on the release.

This is drift, not a regression, and it is re-pinned here — not accepted blind.

## Every hunk, attributed

The whole diff is 10 semantic hunks plus 3 `errorCount` lines.

| snapshot | hunk | cause |
| --- | --- | --- |
| elk | `CommonTabs.vue` **+** `40:39 [TS2322]` — `:key="option"` binds an object where Vue needs a `PropertyKey` | #3443 |
| misskey | `MkTabs.vue` **−** `11:11 [TS7006] Parameter 'el' implicitly has an 'any' type` on `:ref="(el) => tabRefs[t.key] = (el as HTMLElement)"` | #3443 |
| misskey | `MkPageHeader.tabs.vue` **−** the same `11:11 [TS7006]` on the same construct | #3443 |
| npmx.dev | `SkillsModal.vue` `[TS2367]` moves `206:75` → `217:18` | #3509/#3512, then #3510 |
| nuxt-ui | `InputTags.vue` `167:26 [TS2345]` → `167:6 [TS2322]` on `@update:model-value` | #3509 |
| nuxt-ui | `PricingPlan.vue` `231:155 [TS2345]` → `231:148 [TS2322]` | #3509 |
| reka-ui | `CalendarSelect.story.vue` `65:32 [TS2345]` → `65:12 [TS2322]` on `@update:model-value` | #3509 |
| reka-ui | the unanchored `ContextMenu*` message block moves from `ComboboxCustom.story.vue` to `ComboboxReactive.story.vue` | #3443/#3509 |
| reka-ui | `PopperArrow.vue` **+** `42:6 [TS2322]` — `:ref="(el: HTMLElement) => …"` against `VNodeRef` | #3443 |

`errorCount`: elk 280 → 281, misskey 39 → 37, npmx.dev 161 → 161,
[...]

PR #3519, opened 2026-07-31
test(patina): re-pin the five app lint snapshots to no-unused-properties' real usage signal

### LINKED ISSUE #3518: App E2E (lint) is red on the v0.308.0 release commit: five app lint snapshots drifted with #3508
`App E2E (lint)` failed alongside `App E2E (check)` on the v0.308.0 release
commit `83343426f`
([run 30602594088](https://github.com/ubugeeei-prod/vize/actions/runs/30602594088),
job `app-e2e (lint)`), and again on v0.307.0 (`0409c6ade`, run 30599680680).

#3515 re-pins the five **check** snapshots. It deliberately does not touch lint,
so **the release is still blocked on this one.**

## The five failures

`vize lint does not crash and snapshot matches` for `elk`, `misskey`,
`npmx.dev`, `nuxt-ui` and `vuefes-2025`
(`tests/snapshots/lint/__snapshots__/*-lint.snap`).

## The drift is legitimate and is entirely one rule

I regenerated all seven lint snapshots in the App E2E `lint` environment itself
(dispatch on a throwaway branch, snapshot writing forced). Grouping the diff by
`ruleId`:

| snapshot | added | removed |
| --- | --- | --- |
| elk | 4 × `vue/no-unused-properties` | — |
| misskey | 2 × `vue/no-unused-properties` | — |
| npmx.dev | 2 × `vue/no-unused-properties` | 1 × `vue/no-unused-properties` |
| nuxt-ui | 1 × `vue/no-unused-properties` | — |
| vuefes-2025 | 2 × `vue/no-unused-properties` | — |

No other rule moves in any of the five, and `ant-design-vue-lint.snap` and
`reka-ui-lint.snap` were rewritten from scratch in the same run and came back
byte-identical — so the environment is deterministic and the change is confined
to that one rule.

That is #3508, `fix(patina): give no-unused-properties a real template-usage
signal`, doing exactly what it says: the rule now sees real template usage, so
props that only *looked* used are correctly reported, and at least one that
looked unused is correctly no longer reported (the npmx.dev removal).

Same structural cause as #3516: the `lint` suite only runs on
tag/nightly/dispatch, so #3508 was green on its own PR.

## To close this

Re-pin the five snapshots and check the individual diagnostics against their
sources the way #3515 does for check — in particular the one **removal** in
`npmx.dev`, which is the only place a diagnostic disappears.

The regenerated files are already available as the `regenerated-snapshots-lint`
artifact of run 30605114449 if that is useful; they should be re-verified rather
than trusted, since a `--ref` dispatch reflects that branch, not `main`.

ubugeeei:
Second half of the v0.308.0 release unblock. #3515 re-pinned the five **check**
snapshots; this does the same for the five **lint** snapshots, which failed in
the same run and are still red on `main`.

Closes #3518.

## The failure

`App E2E (lint)` failed on `83343426f`
([run 30602594088](https://github.com/ubugeeei-prod/vize/actions/runs/30602594088))
for elk, misskey, npmx.dev, nuxt-ui and vuefes-2025 — `vize lint does not crash
and snapshot matches`.

Same structural cause as #3515 and #3516: `App E2E` `lint` only runs on
tag / nightly / dispatch, so #3508 (`fix(patina): give no-unused-properties a
real template-usage signal`) was green on its own PR and the drift surfaced at
release time.

## Every hunk, attributed

The drift is **entirely `vue/no-unused-properties`**: twelve message objects
change in total — eleven added, one removed — spread over the nine files below.
Each was checked against the fixture source at its pinned submodule commit.

Verified strictly: comparing the global multiset of **full** message objects
(every field, including `ruleDocsPath` and `help`) per snapshot, keyed by file,
the only differences are those twelve. No other rule changes count anywhere, no
retained message moves or is reworded, and the file-entry list is identical in
all seven snapshots.

| snapshot | file | prop | verdict |
| --- | --- | --- | --- |
| elk | `main/MainContent.vue` | `noOverflowHidden` | **+** true positive |
| elk | `publish/PublishAttachment.vue` | `alt` | **+** true positive |
| elk | `publish/PublishWidget.vue` | `inReplyToVisibility` | **+** true positive |
| elk | `status/StatusContent.vue` | `inNotification` | **+** true positive |
| misskey | `MkPushNotificationAllowButton.vue` | `danger`, `link` | **+** true positive |
| npmx.dev | `Package/SkillsCard.vue` | `packageName`, `version` | **+** true positive |
| npmx.dev | `BuildEnvironment.vue` | `buildInfo` | **−** false positive removed |
| nuxt-ui | `prose/Script.vue` | `src` | **+** true positive |
| vuefes-2025 | `SponsorGrid.vue` | `gapSp`, `linkPath` | **+** true positive |

For every added finding, the prop name occurs **exactly once** in the file — in
the `defineProps<{…}>()` type literal — and is not among the names the component
destructures. `nuxt-ui`'s `prose/Script.vue` is the clearest case; that is the
entire component:

```vue
<script setup lang="ts">
defineProps<{
  src: string
}>()

const isDev = import.meta.dev
</script>

<template>
  <div v-if="isDev">
    Rendering the <code>script</code> element is dangerous and is disabled by default. …
  </div>
</template>
```

`vuefes-2025`'s `SponsorGrid.vue` is the one that could have been a trap and is
not — it *does* destructure, but only `imageOnly`, `columns` and `gap`, leaving
`gapSp` and `linkPath` genuinely unreferenced.

### The one removal, in detail

`npmx.dev` `app/components/BuildEnvironment.vue` is the only place a finding
disappears, so it got the same scrutiny as misskey's two vanishing `TS7006`s in
#3515. It is a false positive going away.

```ts
const { footer = false, buildInfo: buildInfoProp } = defineProps<{
  footer?: boolean
  buildInfo?: BuildInfo
}>()

const appConfig = useAppConfig()
const buildInfo = computed(() => buildInfoProp || appConfig.buildInfo)
```

The prop `buildInfo` is destructured **under a different name** and used on the
line below as `buildInfoProp`. Separately, `buildInfo` is rebound as a local
`computed`, and that is what the template reads. The old name-based signal did
not follow the rename, saw no use of a binding called `buildInfo` that came from
the props, and reported it unused. #3508 resolves the aliased binding, so the
warning correctly stops firing — the prop is used.

That the same PR both *adds* findings for genuinely dead props and *removes* one
for an aliased-but-used prop is the signature of a rule that stopped guessing
from names and started reading real usage, which is exactly what #3508 says it
does.

## How these were generated

Re-run with snapshot writing forced on a throwaway branch based on **this exact
`main` (`60754bd38`)** — not the earlier probe branch — inside the App E2E `lint`
environment: same Linux runner, same pinned fixture commits, same misskey
worktree setup. Probe branch deleted.

**Determinism.** The run rewrote all seven lint snapshots from scratch. Every one
of the seven is byte-identical to the same file produced by an independent
earlier run on a *different* base commit — 2.9 MB reproduced exactly, run to run.
Of those seven, `ant-design-vue-lint.snap` (580 KB) and `reka-ui-lint.snap`
(292 KB) also come back byte-identical to what is committed, so exactly the five
expected files move and nothing else drifts.

## Verification

`App E2E` dispatched with `suite=lint` against this branch — the same job that
failed on the release commit.

Local gates: `nix develop -c vp run fmt:check`, `check:js`, `source:lengths` all
exit 0.

## Noted, not fixed here

Every new `vue/no-unused-properties` message anchors at the `<template>` tag
rather than at the prop declaration — `MainContent.vue` reports `41:11`,
`SkillsCard.vue` `13:11`, `prose/Script.vue` `9:11`, and in each case that line
is the file's `<template>` and column 11 is the character just past the
ten-character `<template>` tag. The findings are correct; only the location is
unhelpful. Filed separately rather than widened into a release-unblocking PR.

### DIFF
--- a/tests/snapshots/lint/__snapshots__/elk-lint.snap
+++ b/tests/snapshots/lint/__snapshots__/elk-lint.snap
@@ -1052,6 +1052,17 @@
     "file": "app/components/main/MainContent.vue",
     "messages": [
       {
+        "ruleId": "vue/no-unused-properties",
+        "ruleDocsPath": "docs/content/rules/vue.md",
+        "severity": 1,
+        "message": "[vize:vue/no-unused-properties] Prop 'noOverflowHidden' is defined but never used",
+        "line": 41,
+        "column": 11,
+        "endLine": 41,
+        "endColumn": 11,
+        "help": "Remove unused prop or use it in your template/script"
+      },
+      {
         "ruleId": "html/deprecated-attr",
         "ruleDocsPath": "docs/content/rules/html.md",
         "severity": 1,
@@ -1064,7 +1075,7 @@
       }
     ],
     "errorCount": 0,
-    "warningCount": 1
+    "warningCount": 2
   },
   {
     "file": "app/components/main/MainTitle.vue",
@@ -1873,6 +1884,17 @@
     "file": "app/components/publish/PublishAttachment.vue",
     "messages": [
       {
+        "ruleId": "vue/no-unused-properties",
+        "ruleDocsPath": "docs/content/rules/vue.md",
+        "severity": 1,
+        "message": "[vize:vue/no-unused-properties] Prop 'alt' is defined but never used",
+        "line": 28,
+        "column": 11,
+        "endLine": 28,
+        "endColumn": 11,
+        "help": "Remove unused prop or use it in your template/script"
+      },
+      {
         "ruleId": "a11y/click-events-have-key-events",
         "ruleDocsPath": "docs/content/rules/accessibility.md",
         "severity": 1,
@@ -1907,7 +1929,7 @@
       }
     ],
     "errorCount": 0,
-    "warningCount": 3
+    "warningCount": 4
   },
   {
     "file": "app/components/publish/PublishCharacterCounter.vue",
@@ -1987,6 +2009,17 @@
         "endLine": 457,
         "endColumn": 42,
         "help": "Element IDs must be unique within a document. Duplicate IDs break label associations, ARIA references, and getElementById()."
+      },
+      {
+        "ruleId": "vue/no-unused-properties",
+        "ruleDocsPath": "docs/content/rules/vue.md",
+        "severity": 1,
+        "message": "[vize:vue/no-unused-properties] Prop 'inReplyToVisibility' is defined but never used",
+        "line": 394,
+        "column": 11,
+        "endLine": 394,
+        "endColumn": 11,
+        "help": "Remove unused prop or use it in your template/script"
       },
       {
         "ruleId": "html/deprecated-attr",
@@ -2141,7 +2174,7 @@
       }
     ],
     "errorCount": 4,
-    "warningCount": 11
+    "warningCount": 12
   },
   {
     "file": "app/components/publish/PublishWidgetFull.client.vue",
@@ -2746,9 +2779,21 @@
   },
   {
     "file": "app/components/status/StatusContent.vue",
-    "messages": [],
+    "messages": [
+      {
+        "ruleId": "vue/no-unused-properties",
+        "ruleDocsPath": "docs/content/rules/vue.md",
+        "severity": 1,
+        "message": "[vize:vue/no-unused-properties] Prop 'inNotification' is defined but never used",
+        "line": 37,
+        "column": 11,
+        "endLine": 37,
+        "endColumn": 11,
+        "help": "Remove unused prop or use it in your template/script"
+      }
+    ],
     "errorCount": 0,
-    "warningCount": 0
+    "warningCount": 1
   },
   {
     "file": "app/components/status/StatusDetails.vue",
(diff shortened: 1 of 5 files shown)


ubugeeei:
Verification: `App E2E` dispatched with `suite=lint` against this branch — [run 30607288699](https://github.com/ubugeeei-prod/vize/actions/runs/30607288699), `app-e2e (lint)` **success**, `tests 15 / pass 15 / fail 0`, with `Snapshot matched` for all seven lint snapshots. The same job failed 5/15 on the release commit.

Follow-up filed rather than bundled: #3520 — every `vue/no-unused-properties` message anchors at the `<template>` tag instead of the prop declaration (8/8 of the new messages land on the file's `<template>` line at column 11). The findings are correct; only the reported location is unhelpful. The previous anchor was wrong too, just inconsistently — the removed `buildInfo` message pointed at `16:95-17:30`, which is a `class` attribute inside that file's template.

PR #3713, opened 2026-08-02
fix(fixtures): pin the real-project LSP Corsa binary

ubugeeei:
Shard 10 of Real Project Matrix @ de3243b failed in `Check real-project LSP lifecycle` with `misskey: Timed out waiting for notification textDocument/publishDiagnostics` after burning the full 120s diagnostics wait. The captured server log shows why:

```
WARN vize_maestro::server::state::corsa: corsa bridge spawn failed: Failed to spawn Corsa:
     Failed to start Corsa API session: process is closed: jsonrpc reader
     (working_dir=Some(".../tests/_fixtures/_git/misskey"), corsa_path=None)
flag provided but not defined: -async
```

misskey is its shard's typecheck-performance project, so its dependencies are installed, and it pins `@typescript/native-preview` `7.0.0-dev.20260116.1` while this checkout pins `7.0.0-dev.20260602.1`. The server discovers Corsa by walking the workspace root's ancestors, finds misskey's older binary first, and that build's `api` subcommand has no `-async`. The bridge reports "not available", so the broken probe never gets its TS2322 and the audit waits out the timeout instead of measuring vize.

The fix pins the binary the checkout ships for the audit's sessions, which the resolver honors ahead of its ancestor walk:

```ts
function createPinnedSession(): LspAuditSession {
  pinnedCorsaPath ??= resolvePinnedCorsaPath();
  return new LspSession({ env: { CORSA_PATH: pinnedCorsaPath } });
}
```

`resolvePinnedCorsaPath` mirrors the resolution the editor-e2e harness already does for the same reason (`tools/editor-e2e/real-vue-workspace.mjs`): read `@typescript/native-preview` from the repo root, then the platform package's `lib/tsgo`, probing both glibc and musl on Linux and failing with the paths it checked. `LspSession` now takes an optional `env` layered over the runner's environment, so no other suite's discovery changes.

The four contract tests in `tests/tooling/real-project-lsp.test.ts` still pass; they drive a fake session, so the pinned path is only resolved on the real-workspace path.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/3713"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved language-service audit reliability by consistently using the pinned TypeScript native binary.
  * Added platform-aware binary resolution for Linux environments, including glibc and musl variants.
  * Enhanced diagnostics when the required binary cannot be located.
  * Added support for custom environment settings when launching language-service sessions.
  * Reused the resolved binary across audit sessions for more consistent diagnostics and faster execution.

### DIFF
--- a/tests/tooling/support/lsp/launch.ts
+++ b/tests/tooling/support/lsp/launch.ts
@@ -1,4 +1,6 @@
 import { spawnSync } from "node:child_process";
+import fs from "node:fs";
+import { createRequire } from "node:module";
 import path from "node:path";
 import { root } from "./paths.ts";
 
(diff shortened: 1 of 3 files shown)


ubugeeei:
Superseded by the smaller workflow pin in #3714, which passed the complete 133-project, 11-shard matrix on its final head and is now merged. The user-facing runtime collision remains tracked separately in #3715.

PR #3735, opened 2026-08-02
fix(release): survive main moving under a release in flight

### LINKED ISSUE #3732: fix(release): the release process cannot survive the repository's own merge automation
## The problem

The release process requires `main` to stand still for the duration of the preflight, while the
repository's own automation merges PRs continuously. Those two facts are incompatible, and the
result is that cutting a release is closer to a coin flip than a procedure.

`tools/github/release-preflight.mjs` ends with `verifyGitReleaseTarget(tag, sha)`, which calls
`assertReleaseCommitIsCurrentMain(sha, mainSha)`. Before it gets there, `bootstrapRequiredWorkflowRuns`
dispatches Benchmark, Native Smoke, Fuzz, App E2E and Real Project Matrix at the tag and waits for
all eight required workflows (adding Check, Miri, Deploy docs) to report success at the exact release
SHA — a wait budgeted at 105 minutes and observed to take 30-40. **Any merge to `main` inside that
window invalidates the release**, and because published tags are immutable and the release script
rejects `--force-tag` ("published release tags are immutable"), the only recovery is to burn the
version number and cut the next one.

## Observed record

Seven consecutive attempts, one success:

| tag | outcome | cause |
| --- | --- | --- |
| v0.312.0 | failed | `main` moved during the window |
| v0.313.0 | failed | `Real Project Matrix: completed/failure` |
| v0.313.1 | failed | `App E2E: completed/failure` |
| v0.314.0 | failed | not determined from the job log |
| v0.315.0 | failed | not determined from the job log |
| v0.315.1 | **succeeded** | — |
| v0.316.0 | failed | `main` moved during the window (`2be269255` → `7302f7215`, PR #3731) |

Two distinct failure modes, both structural rather than accidental:

**1. `main` drift.** The maintainer's Codesmith bot merges PRs autonomously within minutes of them
going green. There is no merge freeze, no queue, and nothing that tells the bot a release is in
flight. A release therefore succeeds only if no PR happens to go green during a ~35-minute window.

**2. Flaky required gates.** `App E2E` and `Real Project Matrix` each killed a release. Separately,
`Benchmark` failed v0.312.0's gate set with
`Benchmark regression budget failed for 1 task(s): - Lint (max): 1.054x (+5.37%)` against a 5%
budget — measured on the `chore: release` commit, which changes only version strings in manifests,
`Cargo.lock`, `editors/zed/*` and READMEs and therefore cannot alter lint throughput. The `max`
lanes were introduced in #3598 and their own author measured ~92% of each lane's wall time as
serial, i.e. roughly 12x more variance-sensitive than the thread-pinned lanes. A gate whose noise
band straddles its own threshold cannot distinguish a regression from nothing happening.

## What a fix should establish

Not prescriptions — the right design is the maintainer's call — but any fix has to answer these:

- **A release must be able to acquire exclusive right to `main`**, or stop depending on `main`'s tip.
  Options worth weighing: a merge queue; a repository-level "release in progress" flag the bot and
  `--auto` merges respect; or relaxing `assertReleaseCommitIsCurrentMain` to accept a tag that is an
  ancestor of `main` when the intervening commits are verified not to touch published artifacts.
  The last one is the smallest change and removes the whole failure mode, but it needs a real answer
  for what "does not affect the release" means.
- **A required gate must not fail on a no-op commit.** Either exempt version-bump-only commits from
  the throughput budget, size the `max` lanes' budget from their measured variance instead of
  reusing the 5% figure calibrated for pinned lanes, re-measure once before failing, or make the
  `max` lanes report-only until enough history exists to size them. Whatever is chosen must keep a
  genuine 20% regression failing the gate — state explicitly how.
- **`App E2E` and `Real Project Matrix` need their flake rate characterised.** Two releases died on
  them. If they are genuinely flaky, that is its own defect; if they caught real problems, the
  release process should say so rather than presenting as a generic gate failure.

## Verification bar

- A test over `verifyGitReleaseTarget` / `assertReleaseCommitIsCurrentMain` covering the drift case
  and whatever new acceptance rule lands, asserting complete outcomes by full equality.
- For the benchmark half: a test that a manifest-only diff does not fail the budget, and a test that
  a synthetic 20% regression still does.
- Rerun evidence, not just reasoning: the recovery path today is `gh run rerun <id>` for the whole
  run. Note `gh run rerun <id> --failed` re-runs only `pr-benchmark-budget`, which re-reads the
  already-computed artifact and reproduces the identical number — it can never change the verdict.

## Notes

Also worth folding in: `gh issue view`, `gh pr edit` and `gh pr close` are all broken against this
repository (Projects-classic GraphQL deprecation), so release tooling and agents must use
`gh api` directly. And `.github/workflows/check.yml` is 702 lines, already over the 350-line gate,
so any edit to it must be net-neutral or shrinking or `test-scripts` fails.

ubugeeei:
Closes #3732

## First: one premise in the issue is out of date

The issue says the preflight "calls `assertReleaseCommitIsCurrentMain(sha, mainSha)`, ... a strict
`sha !== mainSha` throw". That function no longer exists. #3540 (`fix(release): accept former main
tips in preflight`, merged 2026-07-31 14:01Z) replaced it with
`assertReleaseCommitIsOnMainFirstParent`, which accepts any commit on `main`'s first-parent history.

That is checkable against the record rather than argued:

- The last release actually killed by drift was run
  [30607798766](https://github.com/ubugeeei-prod/vize/actions/runs/30607798766), whose preflight log
  reads `Release commit 0f0f418bee0b16e589d269aee7a068df2cf00791 is not the current origin/main
  375b0fb5b0d04af9014d81ef865d17dba02f3b0a`. That string has not appeared in a preflight log since
  #3540 landed.
- The later failures in the issue's table have different causes in their logs:
  v0.313.0 `- Real Project Matrix: completed/failure`, v0.313.1 `- App E2E: completed/failure`,
  v0.312.0 `- Benchmark: completed/failure`, v0.315.0 `Release @vizejs/nuxt-lint-config to npm`.
- **v0.316.0 and v0.317.0 are both still running as I write this**, with `main` three commits past
  v0.316.0 (`2be269255` -> `7302f7215` -> `83b6953b0` -> `8e9c10e18`) and one past v0.317.0,
  including the bot-authored and bot-merged #3734. Neither preflight has failed the `main` check.

So the drift *symptom* is already gone. What #3540 left behind is an acceptance rule with no content:
"this commit was on `main` at some point" admits everything, including cases that really should stop
a release. The issue asks for "a real answer for what 'does not affect the release' means". This PR
supplies it, and the tests that were never written for either version of the rule.

## Half 1: what drift is safe, and why it survives an actor we do not control

**Chosen: keep the ancestry allowance, and name the single exception.**

Ordinary drift cannot affect a release, and the reason is structural rather than statistical:

- the published artifacts (crates, npm packages, editor extensions, wasm, CLI binaries) are built
  from the tagged tree, and the tag is immutable and re-verified to still resolve to `sha`;
- every required gate is a `head_sha == release SHA` run, selected by `latestRequiredWorkflowRun`. A
  commit landing on `main` afterwards is not in the tree those gates measured, so it cannot make
  their verdicts stale;
- release-blocking issues are evaluated live at preflight time, not at tag time, so a defect found
  during the window still stops the release.

A fix merged after the tag simply ships in the next version. That is what a version *is*.

One kind of drift is not like that: **another release**. A second `chore: release` commit moves the
version line on, and finishing the older release then publishes a lower version after a higher one --
npm's `latest` dist-tag, the crates.io publish order and the "latest" GitHub Release all end up
describing the superseded build. Not hypothetical: v0.316.0 and v0.317.0 were tagged seven minutes
apart and are in flight together right now, and today's rule lets both publish.

The decidable form of that: **the workspace version at `main`'s tip must still be the version being
released.** Only `tools/moon/cmd/release` writes `[workspace.package].version`, and
`release-local-guard.mjs` requires an exact `main` tip before it does, so that field is precisely
"which release owns `main` now". Equal means the drift is ordinary work; different means a newer
release has taken over. `verifyGitReleaseTarget` reads it with
`git show refs/remotes/origin/main:Cargo.toml` -- no extra fetch, no tag enumeration, no commit
message parsing.

**Why this survives an actor we cannot control.** It requires nothing of the bot. It does not ask it
to observe a flag, respect a queue, or delay a merge. The bot may open and merge PRs at any rate,
including inside the window, and the release still passes -- because the release stopped depending on
`main`'s tip for anything except the one property the bot cannot produce. Only the release tool
changes the workspace version, and it only runs when a human cuts a release.

It also fails *fast*: the check runs in `verifyReleaseTarget`, before `bootstrapRequiredWorkflowRuns`,
so a superseded release stops in seconds rather than after 35 minutes of CI.

**Rejected:**

- *A merge queue.* Correct in principle and the largest change available: it serialises every merge in
  the repository to fix a constraint that exists for ~35 minutes per release, needs branch-protection
  and repository settings this PR cannot carry, and adds latency to every PR. It also does not address
  a second release being cut while the first is in flight, which is the failure this PR actually stops.
- *A repository-level "release in progress" flag.* Only works if the merge automation reads it.
  Nothing in this repository shows where that bot's merge decision is made, so it is unproven -- and
  #3734 was authored *and* merged by it inside a window that had no open non-draft PRs when the tag
  was cut. A fix whose correctness depends on an unobservable actor's cooperation is not a fix.
- *Dropping the ancestry requirement entirely.* Would accept a release commit that reached `main` only
  through a merge's second parent, i.e. one that was never a `main` tip. First-parent membership is the
  proof that this commit was, at some moment, exactly what the repository shipped. Keep it.
- *Requiring the drift to touch no published paths.* This is the reading of "does not affect the
  release" the issue text suggests, and it is the wrong one: it rejects every ordinary merge (the
  bot's PRs touch `crates/**`), which is the failure mode we are removing. The release does not
  contain post-tag commits, so their paths are irrelevant.

## Half 2: a required gate that cannot fail on a no-op commit

`lint-max` and `fmt-max` clear the 1000-file corpus in 37-39ms with the Rayon pool unpinned. The 5%
budget was sized (#1411) for lanes taking 120-190ms, so on these two it is under 2ms of wall time --
inside process startup and scheduler jitter on a shared 32-vCPU runner.

Every release commit is a free null measurement: its diff is version strings in manifests,
`Cargo.lock`, `editors/*` and READMEs, and cannot change throughput. Across the seven such
comparisons run since #3622 introduced the paired estimator (v0.313.0 .. v0.317.0), read from the
`pr-benchmark` artifacts:

| lane | base median | change on a diff that changes no code | sd |
| --- | ---: | --- | ---: |
| Format (max) | 38.9ms | -3.93% .. **+4.49%** | 2.58 |
| Lint (max) | 37.1ms | -3.46% .. +0.60% | 1.37 |
| Type check (1T) | 830.2ms | -1.05% .. +3.29% | 1.46 |
| Lint (1T) | 122.5ms | -2.78% .. +1.90% | 1.51 |
| Compile SFC | 189.9ms | -2.55% .. +1.62% | 1.41 |
| Type check (max) | 839.2ms | -2.45% .. +0.68% | 1.08 |
| Format (1T) | 147.4ms | -1.72% .. +1.23% | 1.20 |

`Format (max)` reached **+4.49%** on v0.315.1's release commit
([run 30713745830](https://github.com/ubugeeei-prod/vize/actions/runs/30713745830)) -- half a point
short of cancelling a release for a commit that changed no code.

On v0.312.0's `+5.37%`: that verdict predates #3622. Rescoring that comparison's own samples with
today's paired estimator gives **-3.93%**, so the specific number in the issue is already fixed. The
lane's noise band still sitting inside the budget is the part that is not.

**Chosen: size these two lanes' budget from that measured distribution -- 10%.** That is 2.2x the
widest excursion ever measured on a no-code diff (4.49%) and more than three standard deviations
above the noisier lane's distribution. `check-max` is unpinned too but keeps 5%: at 839ms it is
nowhere near the noise floor and the same seven comparisons put it inside +-2.45%. The pinned lanes
are untouched.

**A genuine 20% regression still fails, explicitly:** 20% clears the 10% budget by 10 points. To pass,
the pooled paired median would have to be dragged 10 points below the truth; the largest excursion
ever observed on these lanes is 4.49 points and their sd is at most 2.58, so that is roughly a 4-sigma
event -- and #3622's confirmation protocol re-measures a breach and judges the pooled 20 pairs, which
narrows the distribution further. The regressions these lanes were added for (#3460: published
`vize lint (max)` +16%, `vize fmt (max)` +58%) also still fail. `benchmark-lane-budget.test.ts`
asserts the 20% case end to end, through confirmation, down to the exact enforcement message.

**Rejected:**

- *Exempting version-bump-only commits.* Only helps release commits; PRs keep flaking on the same
  lanes. "Provably version-only" also needs content inspection of `Cargo.toml`/`Cargo.lock`/
  `package.json` diffs -- more machinery and more ways to be wrong than a measured budget.
- *Re-measuring once before failing.* Already landed as #3622 and kept; it is why the numbers above
  are as tight as they are. It cannot rescue a lane whose noise band straddles its own threshold --
  pooling only shrinks sd by sqrt(2).
- *Making the max lanes report-only.* Discards exactly the protection #3598 added. #3460's +16% and
  +58% were invisible to the pinned lanes; a report-only lane makes them invisible again.
- *Growing the corpus so the max lanes are slower.* One corpus is shared by all lanes, so this
  lengthens every lane and the whole gate to fix two of them.

## Also investigated, not fixed here

`App E2E` (v0.313.1, [run 30703020598](https://github.com/ubugeeei-prod/vize/actions/runs/30703020598),
job `app-e2e (build)`) and `Real Project Matrix` (v0.313.0,
[run 30700978696](https://github.com/ubugeeei-prod/vize/actions/runs/30700978696), job
`real projects (8/11)`) each killed a release. Their characterisation follows in a comment on this
PR rather than being folded in here: a flaky gate and a mis-sized budget are different defects.

## Tests

- `tests/tooling/release-preflight-runner.test.ts` -- drives the real `--target-only` CLI over a fake
  `git`, asserting `[exitCode, stderr]` by full equality for: exact `main` tip, drift with the version
  unchanged (accepted), drift that bumped the workspace version (rejected as superseded), HEAD/event-SHA
  mismatch, ancestry only through a merge's second parent, and a tag pointing elsewhere. The three
  pre-existing `assert.match` checks are converted to full equality.
- `tests/tooling/release-preflight-core.test.ts` -- `assertReleaseVersionStillOwnsMain` accepts equal
  versions and rejects a superseded one, asserted against the complete `Error`; the first-parent
  rejection is likewise converted from a regex to the complete `Error`.
- `tests/tooling/benchmark-lane-budget.test.ts` (new) -- the lane catalogue's resolved budgets by full
  equality; a real manifest-only sample (v0.315.1's `fmt-max` run times, +4.49%) passing, with the
  whole budget object and enforcement result asserted; the +5.37% rate that cancelled v0.312.0 passing
  on its own lane while still failing on a pinned lane; a 20% regression failing through confirmation,
  down to the exact `Benchmark regression budget failed for 1 task(s) after paired confirmation:` /
  `- Lint (max): 1.200x (+20.00%) over a 10% budget`; and the report naming every lane judged against
  a non-default budget.

No `.contains()`/`.includes()`/substring/regex-on-a-known-string in any new assertion. No workspace
outside `os.tmpdir()`, no build-artifact checks, and `.github/workflows/check.yml` is untouched.

## Commands run

| command | result |
| --- | --- |
| `node --test` over the benchmark, release-preflight, npm-bootstrap, github-workflows, criterion and check-bench tooling tests | 98 tests, 96 pass, 0 fail |
| `nix develop -c node --test tests/tooling/source-file-lengths.test.ts` | 7 pass, 0 fail (after `nix develop -c moon update`) |
| `nix develop -c ./node_modules/.bin/vp check --fix` | formatting clean; no warnings or lint errors across 1389 files |

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/3735"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

### DIFF
--- a/bench/compare-pr-lanes.mjs
+++ b/bench/compare-pr-lanes.mjs
@@ -31,6 +31,50 @@ export function laneEnvironment(threads, base) {
   return env;
 }
 
+/**
+ * Budget for the two lanes whose measured noise is wider than the shared 5%.
+ *
+ * The 5% default was sized (#1411) for lanes that take 120-190ms on the
+ * 1000-file corpus. `lint-max` and `fmt-max` clear that same corpus in 37-39ms
+ * with the Rayon pool unpinned, so 5% of their wall time is under 2ms -- inside
+ * process startup and scheduler jitter on a shared 32-vCPU runner. Every
+ * release commit gives a free null measurement, because its diff is version
+ * strings in manifests, `Cargo.lock`, `editors/*` and READMEs and cannot change
+ * throughput. Over the seven such comparisons run since #3622 introduced the
+ * paired estimator (v0.313.0 through v0.317.0):
+ *
+ * | lane             | base median | change on a no-code diff | sd    |
+ * | ---------------- | ----------: | ------------------------ | ----: |
+ * | Format (max)     |      38.9ms | -3.93% .. +4.49%         |  2.58 |
+ * | Lint (max)       |      37.1ms | -3.46% .. +0.60%         |  1.37 |
+ * | Type check (1T)  |     830.2ms | -1.05% .. +3.29%         |  1.46 |
+ * | Lint (1T)        |     122.5ms | -2.78% .. +1.90%         |  1.51 |
+ * | Compile SFC      |     189.9ms | -2.55% .. +1.62%         |  1.41 |
+ * | Type check (max) |     839.2ms | -2.45% .. +0.68%         |  1.08 |
+ * | Format (1T)      |     147.4ms | -1.72% .. +1.23%         |  1.20 |
+ *
+ * `Format (max)` reached +4.49% on a commit that changed no code: 0.5 points
+ * short of cancelling a release. (v0.312.0's gate did cancel one, reporting
+ * +5.37%, but that predates #3622 -- rescoring that run's own samples with the
+ * paired estimator gives -3.93%.) A budget sitting inside its lane's noise band
+ * reports coin flips, not regressions.
+ *
+ * 10% is 2.2x the widest excursion measured on a diff that changes no code, and
+ * more than three standard deviations above the noisier of the two lanes. It
+ * stays far below what these lanes exist to catch: #3460's published
+ * `vize lint (max)` moved +16% and `vize fmt (max)` +58% over a window the
+ * pinned lanes scored as neutral-to-better, and both still fail at 10%.
+ *
+ * `check-max` is unpinned too but keeps the 5% default: at 839ms it is nowhere
+ * near the noise floor, and the same seven comparisons put it inside +-2.45%.
+ */
+export const UNPINNED_FAST_LANE_THRESHOLD_PERCENT = 10;
+
+/** The regression budget one lane is judged against. */
+export function laneThresholdPercent(task, defaultThresholdPercent) {
+  return task.thresholdPercent ?? defaultThresholdPercent;
+}
+
 export function makeTasks(inputDir, taskFilter) {
   const tsconfig = join(inputDir, "tsconfig.json");
   const pattern = ".";
@@ -58,6 +102,7 @@ export function makeTasks(inputDir, taskFilter) {
       label: "Lint (max)",
       args: ["lint", pattern, "--quiet"],
       threads: "max",
+      thresholdPercent: UNPINNED_FAST_LANE_THRESHOLD_PERCENT,
       allowNonZeroExit: true,
     },
     {
@@ -83,6 +128,7 @@ export function makeTasks(inputDir, taskFilter) {
       label: "Format (max)",
       args: ["fmt", "*.vue", "--check"],
       threads: "max",
+      thresholdPercent: UNPINNED_FAST_LANE_THRESHOLD_PERCENT,
       allowNonZeroExit: true,
     },
     {
--- a/bench/compare-pr-results.mjs
+++ b/bench/compare-pr-results.mjs
@@ -1,6 +1,32 @@
 /**
- * Pure result classification and confirmation for the PR benchmark gate.
+ * Pure result classification, formatting, and confirmation for the PR
+ * benchmark gate.
  */
+
+export function formatRate(value) {
+  if (!Number.isFinite(value)) {
+    return "n/a";
+  }
+  return `${value.toFixed(3)}x`;
+}
+
+export function formatPercent(value) {
+  if (!Number.isFinite(value)) {
+    return "n/a";
+  }
+  const sign = value > 0 ? "+" : "";
+  return `${sign}${value.toFixed(2)}%`;
+}
+
+/**
+ * One budget failure, naming the budget it broke. Lanes no longer share a
+ * single threshold, so a bare rate does not say what the gate compared against.
+ */
+export function formatRegressionLine(regression) {
+  const budget =
+    regression.thresholdPercent == null ? "" : ` over a ${regression.thresholdPercent}% budget`;
+  return `- ${regression.label}: ${formatRate(regression.rate)} (${formatPercent(regression.changePercent)})${budget}`;
+}
 
 function median(values) {
   const sorted = [...values].sort((a, b) => a - b);
@@ -37,6 +63,7 @@ export function summarizeBenchmarkRuns({ id, label, baseRuns, headRuns, threshol
     rate,
     changePercent,
     status,
+    thresholdPercent,
     baseRuns,
     headRuns,
     pairRates,
@@ -58,7 +85,9 @@ export function confirmRegressions(measurements, remeasure, thresholdPercent) {
       label: result.label,
       baseRuns: [...result.baseRuns, ...confirmation.baseRuns],
       headRuns: [...result.headRuns, ...confirmation.headRuns],
-      thresholdPercent,
+      // A lane is judged against its own budget, so pooling the confirmation
+      // samples must not silently re-judge it against the shared default.
+      thresholdPercent: result.thresholdPercent ?? thresholdPercent,
     });
     return {
       ...combined,
--- a/bench/compare-pr.mjs
+++ b/bench/compare-pr.mjs
@@ -12,8 +12,14 @@ import { basename, delimiter, dirname, join, parse, resolve } from "node:path";
 import { performance } from "node:perf_hooks";
 import { pathToFileURL } from "node:url";
 
-import { laneEnvironment, makeTasks } from "./compare-pr-lanes.mjs";
-import { confirmRegressions, summarizeBenchmarkRuns } from "./compare-pr-results.mjs";
+import { laneEnvironment, laneThresholdPercent, makeTasks } from "./compare-pr-lanes.mjs";
+import {
+  confirmRegressions,
+  formatPercent,
+  formatRate,
+  formatRegressionLine,
+  summarizeBenchmarkRuns,
+} from "./compare-pr-results.mjs";
 
 const DEFAULT_RUNS = 5;
 const DEFAULT_WARMUPS = 1;
@@ -71,23 +77,8 @@ function formatMs(ms) {
   })}ms`;
 }
 
-function formatRate(value) {
-  if (!Number.isFinite(value)) {
-    return "n/a";
-  }
-  return `${value.toFixed(3)}x`;
-}
-
 function formatRunList(values) {
   return values.map(formatMs).join(", ");
-}
-
-function formatPercent(value) {
-  if (!Number.isFinite(value)) {
-    return "n/a";
-  }
-  const sign = value > 0 ? "+" : "";
-  return `${sign}${value.toFixed(2)}%`;
 }
 
 function pathWithNodeBins(cwd) {
@@ -159,7 +150,7 @@ function measureTask(task, baseBin, headBin, options) {
     label: task.label,
     baseRuns,
     headRuns,
-    thresholdPercent: options.thresholdPercent,
+    thresholdPercent: laneThresholdPercent(task, options.thresholdPercent),
   });
 }
 
@@ -171,6 +162,7 @@ export function createBenchmarkBudget(results) {
       label: result.label,
       rate: result.rate,
       changePercent: result.changePercent,
+      thresholdPercent: result.thresholdPercent,
     }));
 
   return {
@@ -178,6 +170,20 @@ export function createBenchmarkBudget(results) {
     regressionCount: regressions.length,
     regressions,
   };
+}
+
+/**
+ * Lanes judged against something other than the shared default, so a report
+ * never implies one budget covered every row. Derived from the results rather
+ * than restated, so it cannot drift from what actually gated the run.
+ */
+export function laneBudgetOverrides(results, defaultThresholdPercent) {
+  return results
+    .filter(
+      (result) =>
+        result.thresholdPercent != null && result.thresholdPercent !== defaultThresholdPercent,
+    )
+    .map((result) => `${result.label} ${result.thresholdPercent}%`);
 }
 
 export function renderMarkdown(data) {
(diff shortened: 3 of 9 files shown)


ubugeeei:
## `App E2E` and `Real Project Matrix`: characterised, no new issue warranted

Both gates that killed a release are root-caused, and **both already have landed fixes**. Neither is
a distinct open defect, so nothing is filed and nothing is folded into this PR.

### `App E2E` — v0.313.1, [run 30703020598](https://github.com/ubugeeei-prod/vize/actions/runs/30703020598), `app-e2e (build)`: flake

Failing step `Run selected app e2e suite` (`vp run --filter './tests' test:build`):

```
Built: src/pages/settings/avatar-decoration.dialog.vue -> .../misskey-bu✗ Failed to spawn process: Resource temporarily unavailable (os error 11)
```

`os error 11` is `EAGAIN`: the kernel refused a `fork`. No assertion, no diff, no product output
mismatch, and the message is spliced mid-line into the CLI's own stdout, so there were concurrent
writers. Attempt 1 of the same run failed identically but at a *different* file
(`src/ui/_common_/navbar-h.vue`) — same mode, nondeterministic location, which is resource
exhaustion rather than a deterministic bug. `Failed to spawn process` does not exist anywhere in the
tree at `c3f515b`; it comes from the task runner.

Cause: at that commit `tests/package.json`'s `test:build` ran
`node --test snapshots/build/{generic,elk,misskey,npmx,vuefes}.ts` with no `--test-concurrency=1`,
so Node spawned one child per file up to the CPU count and each child ran `vize build`, which itself
fans out over rayon. Five concurrent full-width builds over the misskey corpus exhausted the runner's
process limit. Fixed by **#3671** (`78e62f58f`, 2026-08-01 15:03Z), which adds
`--test-concurrency=1` — landed ~20 minutes *after* this run and not an ancestor of `c3f515b`. The
same error hit [run 30607730629](https://github.com/ubugeeei-prod/vize/actions/runs/30607730629) on
Jul 31, so it recurred twice before the fix.

### `Real Project Matrix` — v0.313.0, [run 30700978696](https://github.com/ubugeeei-prod/vize/actions/runs/30700978696), `real projects (8/11)`: real defect, since fixed

Failing step `Check glyph formatter corpus properties`:

```
✖ glyph corpus idempotence holds for every hydrated fixture
  AssertionError: 1 idempotence violation(s):
  crater: resources/scripts/components/base/BaseGlobalLoader.vue
  first divergence at line 212:
    fmt2 > @keyframes orbit {
```

A deterministic violation of `fmt(fmt(x)) == fmt(x)`; the sibling lint-agreement and
parse-preservation sweeps passed, counts were self-consistent (1326 stable + 1 violation = 1327
files), and the file is not in `tests/_fixtures/glyph-corpus-known-violations.json` (the log confirms
`0 known violation(s) skipped`). Fixed by **#3668** (`20e7a285d`, 2026-08-01 13:56Z), touching only
`crates/vize_glyph/src/style/stabilization.rs`. It was a formatter bug and not corpus drift: the
crater submodule pin is byte-identical across both days
(`05d5ce26fdd8d9466009163444e944259bc0cc2a`), and today's shard-8 run reports
`glyph idempotence: 1327 file(s) ... 0 violation(s)`.

### Measured reliability

`e2e.yml`, 600 runs (Jul 29 06:28Z .. Aug 2 12:04Z): 462 success, 104 cancelled, 32 failure, 2 in
progress — **6.5% of completed runs fail**. Cancellations are ordinary concurrency-group superseding.
Of the 32 failures: `app-readiness` 20, `app-e2e (check)` 8, `app-e2e (lint)` 3, `app-e2e (build)` 2,
and one each for `vrt`/`preview`/`dev`. Only 8 runs used a second attempt, so "failed then passed on
rerun" is not a usable signal here.

`real-project-matrix.yml`, 85 runs (Jul 17 15:39Z .. Aug 2 11:54Z): 61 success, 24 failure, 0
cancelled — **28.2% fail**, and none was ever retried. Failing steps across the 24:
`Exercise real projects with every core tool` 9, `Record typechecker baseline divergence` 6,
`Check glyph formatter corpus properties` 6, `Upload shard report` 2,
`Select and hydrate fixture shard` 2, and one each for the pinned-baseline install, syntax
highlighting and LSP lifecycle steps.

So the two workflows have opposite profiles: `App E2E`'s failures are infrastructure, while `Real
Project Matrix`'s 28% is dominated by genuine product findings. That gate is doing its job; it is
simply finding a lot.

### One thing worth the maintainer's attention

Glyph corpus idempotence is a repeat offender — 6 of the 24 matrix failures — and the earlier
occurrences were *different* bugs from the keyframes one: the Jul 26 and Jul 29 shard-8 runs failed
on `dashy: src/components/Settings/CustomThemeMaker.vue` and `vue-data-ui: src/components/vue-ui-flow.vue`,
both mis-indenting multi-line inline event handlers on the second pass. That is a recurring *class*
of formatter non-idempotence being retired one instance at a time. There is no live violation today
(0 across 1327 files), so there is no defect to file; whether the class deserves a systematic fix
(for example property-based fuzzing of `fmt . fmt` over the corpus) is a call for the maintainer.

Issue #3756, opened 2026-08-02
fix(canon): vize and vue-tsc disagree systematically on TS7006 for template and handler parameters


ubugeeei:
## Summary

Across the production matrix, vize and `vue-tsc` disagree **systematically** on `TS7006`
("Parameter implicitly has an 'any' type") for template and handler parameters: vize under-reports
on 4 projects and over-reports on 2. This is not one project's quirk — it is a consistent behavioural
difference in how the two model the implicit typing of those parameters.

## Precondition — read this before acting

This was surfaced while diagnosing #3738, whose finding is that the divergence baseline was
**instrument-broken on most shards**: `materializeBaselineProject` wrote
`files: [<vize's Vue corpus>], include: []`, and a `files` list seeds the program with those roots
only, so every ambient declaration the fixture owns left the program. The baseline then reported the
project's own globals as undeclared and scored that against vize.

#3755 fixes that. **Do not start on this issue until #3755 has landed and a fresh matrix sweep has
run**, because today's numbers are contaminated. The `TS7006` finding is called out separately from
the instrument problems precisely because it looked structural rather than environmental — implicit
-any on a parameter cannot be caused by a missing ambient declaration — but it still deserves to be
re-measured against a trustworthy baseline before anyone designs a fix.

## Evidence as it stands (pre-#3755, treat as indicative)

From the weekly matrix run [30738583070](https://github.com/ubugeeei-prod/vize/actions/runs/30738583070):

- **reka-ui** — classified 98% real divergence. 23 FN `TS7006`, alongside 55 FP `TS2322`
  `ComputedRef<boolean | undefined>` and 22 FN `TS1117` (purely syntactic, so no ambient type can
  explain them).
- **misskey** — mixed; its real residue includes 10 FN `TS7006` and 22 FP `TS2345` handler
  contravariance.
- Under-reporting on 4 projects, over-reporting on 2 (per the #3738 diagnosis).

The `TS2345` handler-contravariance divergence on misskey is plausibly the same root behaviour seen
from the other side — how a handler's parameter type is derived — and should be investigated
together rather than as two issues.

## What "done" means

1. Re-measure against a post-#3755 sweep and state the real per-project `TS7006` counts.
2. Determine which side is right, per shape, using `vue-tsc` as the oracle. Both directions matter:
   a false negative means vize silently accepts code `vue-tsc` rejects; a false positive means vize
   reports an error on correct code. Neither is acceptable and they may have different causes.
[...]

PR #3823, opened 2026-08-03
fix(canon): surface unknown event callback implicit any

ubugeeei:
## Summary

- preserve `noImplicitAny` diagnostics for inline callbacks on unresolved component events
- keep known emit tuples contextually typed and Vue 2's permissive legacy fallback unchanged
- add an exact TS7006 regression oracle based on Misskey's `MkSuspense @resolved` case

## Root cause

Unresolved component events used `(...args: any[]) => unknown` as the contextual type for authored inline callbacks. That synthetic `any` context suppressed TS7006 diagnostics that vue-tsc reports.

The generated callback initializer now receives `unknown` when event arguments are unresolved. Known events still receive their inferred listener signature; invocation uses the existing callable fallback after the initializer is checked.

## Validation

- `VIZE_TEST_REQUIRE_TSGO=1 cargo test -p vize_canon --lib` (800 passed)
- `cargo clippy -p vize_canon --lib -- -D warnings`
- `cargo fmt --all -- --check`
- exact vue-tsc 3.3.4 / TypeScript 6.0.3 oracle: `src/App.vue(2,27): error TS7006: Parameter 'result' implicitly has an 'any' type.`

Local `source-file-lengths` could not complete because the installed MoonBit 0.1.20260512 rejects the current `moon.mod` `license` key; CI installs the repository-pinned toolchain and remains the acceptance gate.

Refs #3756
Refs #3222

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/3823"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>


ubugeeei:
Superseded by the narrower #3756 follow-up validated against real-project fixtures. Broad unresolved-event contextual typing regresses the Misskey matrix from 3 to 124 Vize-only TS7006 diagnostics, so this approach will not be merged.

PR #3882, opened 2026-08-05
fix(vite-plugin): register SFCs in ssrContext.modules during SSR

### LINKED ISSUE #3868: fix: Nuxt SSR page styles flash on initial load with `@vizejs/nuxt`
### Area

Nuxt integration

### Version

@vizejs/nuxt@0.325.0

### Minimal reproduction

https://github.com/wattanx/vize-nuxt-ssr-style-repro

```sh
pnpm install
pnpm build
node .output/server/index.mjs
```

The reproduction uses a minimal Nuxt pages setup:

```vue
<!-- app/app.vue -->
<template>
  <NuxtPage />
</template>
```

`app/pages/index.vue` contains the rendered markup and a regular
`<style scoped>` block. Inline styles are disabled so the expected production
stylesheet link is directly observable:

```ts
export default defineNuxtConfig({
  modules: ["@vizejs/nuxt"],
  vize: {
    compiler: true,
    lint: false,
  },
});
```

### Actual behavior

On a cold load or hard reload, the server-rendered card first appears unstyled.
After client-side JavaScript loads the route chunk, the background, border,
typography, and layout styles are applied at once. This causes a visible flash
of unstyled content and layout shift during the initial page load.

The CSS asset is generated successfully as
`.output/public/_nuxt/index.*.css`, and the client manifest associates it with
the page. The missing link therefore occurs while Nuxt resolves the modules
used during SSR, rather than during CSS compilation or asset serving.

With JavaScript disabled, the server-rendered page remains unstyled instead of
recovering after the flash.

The same initial style flicker can also be observed with `pnpm dev` in the Vize
project.

### Expected behavior

The page should be correctly styled from the first paint, without a flash or
layout shift when client-side JavaScript starts. It should also remain styled
when JavaScript is disabled.

### Additional investigation

Nuxt's default Vite/Vue pipeline uses `@vitejs/plugin-vue`. During an SSR
transform, the plugin wraps the component's `setup` function and adds the
normalized SFC path to `ssrContext.modules`.

After rendering, `vue-bundle-renderer` merges `ssrContext.modules` with the
entrypoints and resolves their dependencies from the client manifest. Nuxt then
adds the returned styles to the document head.

Implementation links:

- [`@vitejs/plugin-vue`: SSR module registration](https://github.com/vitejs/vite-plugin-vue/blob/d8ff7d0e8f557a7c1975c07b30e232c69bdbbc03/packages/plugin-vue/src/main.ts#L198-L210)
- [`vue-bundle-renderer`: request module IDs](https://github.com/nuxt-contrib/vue-bundle-renderer/blob/e7080f9f9bb9da57b67f198f715b06b99d8d3957/src/runtime.ts#L289-L301)
- [`vue-bundle-renderer`: dependency resolution](https://github.com/nuxt-contrib/vue-bundle-renderer/blob/e7080f9f9bb9da57b67f198f715b06b99d8d3957/src/runtime.ts#L306-L316)
- [Nuxt: request dependency lookup](https://github.com/nuxt/nuxt/blob/57614578c0102c16b69b64f1dac186c732856e90/packages/nitro-server/src/runtime/handlers/renderer.ts#L248-L251)
- [Nuxt: stylesheet link generation](https://github.com/nuxt/nuxt/blob/57614578c0102c16b69b64f1dac186c732856e90/packages/nitro-server/src/runtime/handlers/renderer.ts#L290-L305)

### Environment

- OS: macOS 26.5.2
- Architecture: arm64
- Node.js: v24.18.0
- Package manager: pnpm 10.30.1
- Framework: Nuxt 4.5.0
- Vite: 8.1.5
- Vue: 3.5.35 (resolved as 3.5.40 during verification)
- Vue Router: 5.2.0
- `@vizejs/nuxt`: 0.325.0

### Before submitting

- [x] I checked existing fix reports and pull requests.

ubugeeei:
Closes #3868.

## What

`vue-bundle-renderer` intersects `ssrContext.modules` with the client manifest to decide which stylesheets belong in the document head. Vize never registered its SFCs there, so server-rendered markup arrived unstyled and only gained its styles once the route chunk loaded — a flash that never recovers with JavaScript disabled.

## Why this is a second attempt

#3870 shipped the same registration and had to be reverted in #3872: it broke Nuxt prerendering of the vuefes-2025 app.

The cause, found by bisecting a local reproduction of that app's build: the registration appended its `import` at the **end** of the module. That is legal ESM — imports hoist — but it makes the module's last import statement its last line, and Nuxt's auto-import injection then adds nothing. `SectionStaff.vue` lost its auto-imported `useI18n`, and prerendering `/` failed with

```
ReferenceError: useI18n is not defined
    at setup (SectionStaff.vue-Cc1UHvCP.mjs:336:17)
```

while every sibling route still built — which is why the failure read as route-specific rather than as a transform bug.

This PR splits the emitted snippet: the `import` is prepended, the wrapper is still appended so it closes over whatever `setup` the emitter's own rewrites left in place.

## Verification

Against the real `vuefes-2025` fixture, built locally three times:

| build | result |
| --- | --- |
| without the change (control) | green, no registrations |
| #3870 as merged | **`[500]` prerendering `/`**, `useI18n` import dropped from 1 chunk |
| this PR | green, no dropped imports, **40 modules carry the registration** |

`node src/test.ts` in `npm/builder/vite`: 47/47 pass, including a new test asserting no import may follow the emitted module body.

Note that `app-e2e` — which covers this app — is skipped on `pull_request` and only runs on tags, schedule, or dispatch, so this PR's own checks do not exercise it. That is what let #3870 through.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/3882"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

### DIFF
--- a/npm/builder/vite/src/plugin/load.ts
+++ b/npm/builder/vite/src/plugin/load.ts
@@ -11,6 +11,7 @@ import {
   type VizePluginState,
 } from "./state.ts";
 import { getLoadableVueSfcPath, shouldLoadCompiledVueSfcPath } from "./load-sfc.ts";
+import { appendSsrModuleRegistration, normalizeVueServerRendererImport } from "./ssr-modules.ts";
 import { compileFile, compileJsxModule } from "../compiler.ts";
 import { embedsInlineCss, generateOutputWithMap, hasDelegatedStyles } from "../utils/index.ts";
 import { MappedModule, type SourceMapV3 } from "../utils/source-map.ts";
@@ -31,6 +32,8 @@ import {
   rewriteStaticAssetUrls,
 } from "../transform.ts";
 import { transformVizeVirtualModule } from "./vite-transform.ts";
+
+export { normalizeVueServerRendererImport };
 
 /** What `load` hands Vite: emitted code plus its map, when one was produced. */
 type LoadResult = { code: string; map: SourceMapV3 | null };
@@ -53,10 +56,6 @@ export function getBoundaryPlaceholderCode(realPath: string, ssr: boolean): stri
     return SERVER_PLACEHOLDER_CODE;
   }
   return null;
-}
-
-export function normalizeVueServerRendererImport(code: string): string {
-  return code.replace(/\bfrom\s+(['"])@vue\/server-renderer\1/g, 'from "vue/server-renderer"');
 }
 
 function findMacroArtifactModule(
@@ -153,6 +152,7 @@ function loadCompiledSfcModule(
   rewritten.edit(rewriteDynamicTemplateImports(rewritten.code, state.dynamicImportAliasRules));
   rewritten.edit(rewriteStaticAssetUrls(rewritten.code, state.dynamicImportAliasRules));
   rewritten.edit(rewriteImportMetaGlobBase(rewritten.code, realPath, state.root));
+  rewritten.edit(appendSsrModuleRegistration(rewritten.code, realPath, state.root, isSsr));
   return { code: rewritten.code, map: rewritten.map };
 }
 
--- a/npm/builder/vite/src/plugin/ssr-modules-load.test.ts
+++ b/npm/builder/vite/src/plugin/ssr-modules-load.test.ts
@@ -1,0 +1,75 @@
+import assert from "node:assert/strict";
+import { test } from "node:test";
+
+import type { VizePluginState } from "./state.ts";
+import { loadHook } from "./load.ts";
+import { toVirtualId } from "../virtual.ts";
+
+const ROOT = "/project";
+const PAGE = "/project/app/pages/index.vue";
+
+const COMPILED = {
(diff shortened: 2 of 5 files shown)


ubugeeei:
The full app test also passes locally, not just the raw build:

```
VIZE_TEST_WORKTREE_ID=verify RUN_BUILD_TESTS=1 node --test tests/app/preview/vuefes.ts
✔ vuefes-2025 build (79243ms)
ℹ tests 1  ℹ pass 1  ℹ fail 0
```

That is the test `app-e2e` runs and that #3870 broke.

Issue #3282, opened 2026-07-28
feat(canon): adopt TypeScript content mappers for native Vue support


ubugeeei:
## Context

TypeScript native is designing first-party external file support in [microsoft/typescript-go#4712](https://github.com/microsoft/typescript-go/pull/4712). Vize already owns the SFC parser, virtual TypeScript generator, authored mappings, CLI, LSP, and tsgo/Corsa integration, so the Content Mapper boundary should become the zero-duplication path into TypeScript rather than another compatibility layer.

The upstream PR is still open, but its exact head is testable and its protocol has materially evolved. As of 2026-08-09 the audited head is `bddd2162710e50281fa838456a875fd59ee7c91f`.

## Landed Vize-side foundation

- [x] Add a long-lived `vize content-mapper` stdio process using JSON-RPC 2.0 with LSP `Content-Length` framing.
- [x] Implement protocol v1 `initialize` and `transform` with negotiated UTF-8 positions and `vize` diagnostics.
- [x] Convert Canon virtual-code mappings into `Verbatim` / `Atom` span tuples with non-overlapping generated ranges.
- [x] Return authored Vue parse diagnostics as mapper diagnostics instead of process failures.
- [x] Advertise the mapper from the published `vize` package through `tsContentMapper` metadata.
- [x] Document the experimental `tsconfig.json` opt-in and `--loadExternalPlugins` requirement.
- [x] Activate `TypeScriptTeam.native-preview` and request `.vue` Content Mapper discovery from the Vize VS Code extension.

## Critical protocol drift found in the production-readiness audit

Vize currently defines:

```rust
const MAPPING_PURPOSE_ALL: usize = 3;
```

and writes that value into every mapping tuple. On current upstream, tuple element 6 is `SpanMapFeature`, where `3` means only `Hover | SignatureHelp`; `All` is `(CodeLens << 1) - 1` (currently `2_097_151`). Completion, definition, type definition, implementation, references, rename, code actions, semantic tokens, and most other mapped features are therefore disabled by Vize's purported "all" mask.

The current unit tests pass because they validate Vize's own tuple assumptions, not an exact upstream tsgo build.

Additional drift and design gaps:

- `TransformParams` ignores mapper `options` and effectively ignores `compilerOptions`.
- Canon generation hard-codes `options_api: false`, `legacy_vue2: false`, default dialect/template syntax, and default check options.
- The manifest declares neither compiler-option dependencies nor dynamic configuration identity/watch files.
[...]

PR #3898, opened 2026-08-06
fix(canon): register out-of-root workspace .vue files reachable through imports

### LINKED ISSUE #3887: fix(canon): .vue imports outside the scanned roots silently degrade to the ambient stub
The single biggest monorepo defect found in the clean-room audit. `vize check src` inside a workspace app — the obvious invocation, and what every CI recipe does — silently stops type-checking every component imported from a workspace package.

## Repro

pnpm workspace: `packages/ui` exports `UiButton.vue` (props `variant: "primary" | "ghost"`, emit `press: [variant]`) through `src/index.ts`; `apps/web` consumes it via a `#ui` paths alias. `apps/web/src/components/Panel.vue`:

```vue
<UiButton variant="danger" @press="(v) => v.toFixed(2)" />
<GlobalBtn :variant="label" />  <!-- GlobalComponents-registered, same package -->
```

## vue-tsc (from apps/web)

```
Panel.vue(17,15): error TS2322: Type '"danger"' is not assignable to type '"primary" | "ghost"'.
Panel.vue(17,49): error TS2551: Property 'toFixed' does not exist on type '"primary" | "ghost"'.
Panel.vue(19,17): error TS2322: Type 'string' is not assignable to type '"primary" | "ghost"'.
```

## vize 0.333.0 (`vize check src` from apps/web)

All three **silent**, plus a knock-on false positive:

```
Panel.vue  error:17:41 [TS7006] Parameter 'v' implicitly has an 'any' type.
```

Control experiment proving the root cause: `vize check apps/web/src packages/ui/src` from the workspace root reports all three correctly and the TS7006 disappears. The behavior difference is purely whether `UiButton.vue` is inside the scanned set.

## Root cause

The virtual project only virtualizes `.vue` files in the scanned roots (`BatchTypeChecker::scan_paths`). An import resolving to a real-tree `.vue` outside that set falls back to the ambient `*.vue` stub — `tsconfig_gen.rs` documents this ("`.vue` files only match the ambient `*.vue` stub (default export only)"). The stub component accepts any props and its emit payloads are untyped, so consumer-side checking silently vanishes. vue-tsc type-checks the module graph, not a file list, so it never has this cliff.

## Acceptance

- After registering the explicit scan set, imports are resolved (respecting tsconfig `paths`, workspace `node_modules` symlinks, and relative specifiers) and any `.vue` reachable through them whose canonical path is outside `node_modules` (pnpm workspace symlinks canonicalize outside it) is registered too, transitively, to a fixpoint.
- The monorepo repro reports the three diagnostics above from `apps/web` with only `src` scanned, and the TS7006 is gone.
- Published-package `.vue` files that ship inside real `node_modules` (canonical path stays inside) keep the current stub behavior — that half is #3282's content-mapper territory, not this issue.
- No diagnostic *duplication* when the same dependency file is reachable from several roots, and dependency-file diagnostics map to their own paths (vue-tsc reports errors inside out-of-include imported files as well; matching that is in scope).

ubugeeei:
Closes #3887 — the monorepo killer from the clean-room audit.

## What

`vize check src` from a workspace app silently stopped type-checking every component imported from a sibling package: the import fell back to the ambient `*.vue` stub, props/emits became `any`, and the only trace was a knock-on `TS7006`. Three coordinated pieces fix it end to end:

1. **Reachability fixpoint** (`dependency_scan.rs`) — after `scan_paths`, the import graph of everything registered is walked and reachable first-party files register too. First-party is judged on the *canonical* path: a pnpm workspace symlink canonicalizes out of `node_modules` and registers; published packages keep the stub (#3282's territory); declaration files are refused, since a `.d.ts` holds no component to check and its ambient declarations are program-wide (a script `.d.ts` carrying `declare module "vue"` would replace Vue's real typings). Specifiers come from the generated content (always valid TS) with the rewriter's `.vue.ts` spelling folded back; resolution covers relative specifiers and `paths` aliases anchored by the same flattened chain as the generated tsconfig (#3891).
2. **External escape subtree** (`external_mirror.rs`) — out-of-root files mirror under `__vize_external__/<abs-path-replayed-as-directories>`, so relative imports *between* external files (a mirrored barrel's `./UiButton.vue` beside its generated companion) resolve exactly as in the real tree. Both path mappers escape there instead of erroring; a `..` component is refused so the subtree cannot be escaped.
3. **Alias routing** (`vue_alias.rs`) — an escaped absolute target gains the external-mirror candidate ahead of its real-tree fallback, so rewritten barrels and generated companions win and everything unmirrored falls through unchanged.

## Verification

The audit's monorepo reproduction (pnpm workspace, `#ui` barrel alias, GlobalComponents via env.d.ts), single-root run before → after:

| diagnostic | before | after |
|---|---|---|
| `variant="danger"` prop union (TS2322) | **silent** | reported at vue-tsc's position |
| emit payload `toFixed` (TS2551) | **silent** | reported |
| GlobalComponents `:variant` (TS2322) | **silent** | reported |
| `TS7006` implicit-any knock-on | false positive | gone |

Output is byte-equal to the two-root control run that always worked. The clean-room single-project fixture is unchanged at 23 diagnostics (no new false positives), `cargo test -p vize_canon` is 890/0 including new unit tests for the escape mapping and alias candidates, and a new CLI integration test (`check_workspace_vue_cli.rs`) pins the whole scenario end to end against the real checker.

Scope note, recorded in the module docs: bare workspace-package specifiers (`@scope/pkg` without a `paths` alias) resolve only through their alias today — package-exports resolution is a follow-up. The LSP hover-side of #3887 rides the same virtual project and is verified next.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/3898"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved type checking for Vue components imported from sibling workspace packages.
  - Invalid component props and event payloads are now detected more reliably.
  - Resolved issues affecting dependencies located outside the current project directory.
  - Improved support for TypeScript path aliases and external Vue/TypeScript files.
  - Reduced false implicit-`any` diagnostics when checking cross-package components.
  - Improved dependency discovery through imports, including aliased and workspace-linked files.
  - Improved handling of external files and path mapping across platforms.


ubugeeei:
Superseded: every commit here landed on main through #3899, which stacked on this branch and merged first while this PR's perf guards (dialect-guard / criterion-ab) kept flaking at their 5ms absolute threshold — the identical code passed both guards on #3899's runs, which is what ruled out a real regression.

Issue #3911, opened 2026-08-07
Template expression completion ignores member-access context


ubugeeei:
Round-2 stdio audit on shipped v0.338.0 (fixture: `const users: User[]` with `interface User { id: number; name: string; email?: string }`; template `<li v-for="it in users" :key="it.id">{{ it.name }}</li>`). Both servers driven over real stdio; Volar oracle is `@vue/language-server` 2.2.10 non-hybrid + TS 5.8.3.

`textDocument/completion` right after the `.` in `{{ it.| }}`:

| | items |
|---|---|
| Volar | `id`, `name`, `email?` (3 — the member list, with optionality marker) |
| vize 0.338.0 | `Child`, `users` (2 — the setup-binding list) |

vize answers the identifier-scope list regardless of the member-access context, so typing `it.` in a template suggests unrelated top-level bindings and never the actual properties. The same wrong list comes back at the identifier position before the dot too, where it is at least contextually plausible (Volar there offers its full scope list including lib globals — 1007 items; vize offers only the two setup bindings, a breadth divergence worth its own look but not the correctness problem).

The hover pipeline already resolves this exact position through the canonical virtual document and answers `const it: User` (backend-verified since #3909), so the type information exists — completion just never consults the backend for template expressions.

Acceptance: member completion after `.` in a template expression lists exactly the target type's properties (with optionality reflected), verified over stdio against the Volar oracle on this fixture; identifier-position completion keeps at least the current bindings.

PR #3920, opened 2026-08-07
fix(maestro): member access detection follows identifier characters, not ASCII bytes

ubugeeei:
Follow-up to #3913, which landed `is_at_member_access_position` as the predicate that routes a template member access through the checker. Review on that PR caught two cases its byte scan gets wrong.

The scan stepped back over the partial member name one ASCII byte at a time, so it stopped at the first byte of any multi-byte character. `{{ it.名| }}` therefore looked like a bare identifier, took the structural binding list, and suggested the SFC's bindings instead of `User`'s members: the exact symptom #3911 describes, just for a Unicode member name.

The digit guard was also asking the wrong question. It tested the first character of the partial *name*, which only catches a numeric fragment that still has a digit under the caret (`1.5|`), so `1.|` and `1.na|` both routed to the checker. The receiver is what decides whether the `.` belongs to a number, so the guard now tests the start of that token:

```rust
let dot = pos - 1;
// `...rest` spreads an expression rather than reading a member off it.
if content[..dot].ends_with('.') {
    return false;
}
// A digit-led receiver means the `.` belongs to a numeric literal rather
// than to a member read, which covers every fragment of one (`1.`, `1.5`,
// `1.na`). An identifier that merely contains digits (`foo1.bar`) is still
// a member read, so this tests the start of the receiver token, not the
// character next to the dot.
!content[identifier_start(content, dot)..dot]
    .starts_with(|ch: char| ch.is_ascii_digit())
```

Scanning is now character-based via a shared `identifier_start` helper, used for both the member name and the receiver. `is_identifier_char` is `is_alphanumeric() || '_' || '$'` rather than a `unicode-ident` lookup, since that crate is only a transitive lock entry here and the distinction (a token *starting* with a Unicode digit) cannot occur in either a valid identifier or a numeric literal.

Neither case changes the routing decision for anything that already worked: receivers ending in `]`, `)`, `?`, or `!` still reach the checker, `...rest` and plain identifier positions still take the structural list, and `foo1.bar|` stays a member read. Covered by new assertions for `it.名|`, `it.名前|`, `名.前|`, `foo1.ba|`, `1.|`, `1.na|`, and `42.toStrin|`.

Opened as a separate PR because #3913 was merged (and its branch deleted) before this fix was ready.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/3920"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved member-access detection for Unicode identifiers, including combining marks and zero-width joiners.
  * Added support for whitespace between the dot and a partial member name.
  * Correctly distinguishes numeric fragments from identifiers and completed numeric literals.
  * Preserved support for spread access and member access on completed numeric literals.

* **Tests**
  * Expanded coverage for Unicode names, whitespace, numeric fragments, numeric members, and special numeric literal forms.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3870,6 +3870,7 @@ dependencies = [
  "oxc_diagnostics",
  "oxc_parser",
  "oxc_span",
+ "oxc_syntax",
  "parking_lot",
  "ropey",
  "serde",
--- a/crates/vize_maestro/Cargo.toml
+++ b/crates/vize_maestro/Cargo.toml
@@ -46,6 +46,7 @@ oxc_ast.workspace = true
 oxc_ast_visit.workspace = true
 oxc_parser.workspace = true
 oxc_span.workspace = true
+oxc_syntax.workspace = true
 oxc_diagnostics.workspace = true
 
 # Internal crates
--- a/crates/vize_maestro/src/ide/template_expression.rs
+++ b/crates/vize_maestro/src/ide/template_expression.rs
@@ -98,9 +98,13 @@ fn identifier_start(content: &str, end: usize) -> usize {
     start
 }
 
+/// ECMAScript `IdentifierPart`, so the scan keeps every character an editor can
+/// have already typed into a member name: `is_alphanumeric` alone drops the
+/// combining marks of decomposed text (`café` as `cafe` + U+0301) and the
+/// zero-width joiners, both of which end the token one character too early.
 #[cfg(feature = "native")]
 fn is_identifier_char(ch: char) -> bool {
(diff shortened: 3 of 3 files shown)


ubugeeei:
@codesmith-bot fix ci

PR #3922, opened 2026-08-07
fix(maestro): anchor tsconfig paths on baseUrl and resolve alias barrels

ubugeeei:
Follow-up to #3919 (already merged), addressing three review findings on the alias re-spelling work.

**`paths` targets now anchor on `compilerOptions.baseUrl`.** `project_paths` returned the declaring config's directory unconditionally, but TypeScript resolves `paths` targets against `baseUrl` when it is set. With `baseUrl: "./src"` and `"@/*": ["*"]`, `@/foo` resolved to `<root>/foo` instead of `<root>/src/foo`, so both the definition jump and rename re-spelling missed every alias in that (very common) `create-vue`-shaped config:

```rust
let config_dir = config_path.parent()?;
// TypeScript resolves `paths` targets against `baseUrl` when it is set,
// falling back to the declaring config's directory otherwise.
let anchor = match compiler_options.get("baseUrl").and_then(|v| v.as_str()) {
    Some(base_url) => config_dir.join(base_url),
    None => config_dir.to_path_buf(),
};
```

**Trailing commas no longer void the whole config.** The jsonc reader stripped comments but handed the result to `serde_json`, which rejects the trailing commas `tsc` accepts, so a single stray comma anywhere in `tsconfig.json` silently dropped `paths` entirely: the same failure mode as the comment bug #3919 fixed. The stripper now removes them in the same string-aware pass, treating a closing delimiter as what makes a preceding comma trailing:

```rust
// A closing brace or bracket retroactively makes any comma that
// precedes it (across whitespace and stripped comments) trailing.
'}' | ']' => {
    while out.ends_with(char::is_whitespace) {
        out.pop();
    }
    if out.ends_with(',') {
        out.pop();
    }
    out.push(c);
}
```

**Aliased directory-index imports get rename edits.** `probe_alias_target` probed `base` and `base.<ext>` only, so `@/components/Button` backed by `src/components/Button/index.ts` resolved to nothing and silently received no edit. It now falls through to `base/index.<ext>`, mirroring the relative scanner's candidate order, and carries the relative scanner's `RenderStyle` instead of a bare `extensionless` flag. That lets the barrel spelling survive the move (`@/widgets/Button`, not `@/widgets/Button/index`) while the out-of-subtree fallback reuses `render_module_specifier` verbatim. A target landing directly on the alias root degrades to `@/index`, since `@/` is not a module.

All three are unit-pinned: two anchor tests plus a comments-and-trailing-commas test in `tsconfig_paths.rs`, and a barrel-directory move in `alias_tests.rs`.

Opened as a new PR because #3919 was merged (and its branch deleted) before these comments were addressed.

## Summary by CodeRabbit

- **Bug Fixes**
  - File renaming now preserves explicit, extensionless, and directory-index import styles.
  - Aliased imports retain the correct directory spelling when referenced folders are renamed.
  - Path aliases now resolve relative to the configured base URL when available.

- **Improvements**
  - Improved handling of comments and trailing commas in configuration files.
  - Added coverage for alias-based renaming and configuration path resolution.

### DIFF
--- a/crates/vize_maestro/src/ide/file_rename/alias.rs
+++ b/crates/vize_maestro/src/ide/file_rename/alias.rs
@@ -13,8 +13,9 @@
 use std::path::{Path, PathBuf};
 
 use super::manual::{
-    RESOLVABLE_SCRIPT_EXTENSIONS, RenameTarget, apply_all_path_renames, candidate_exists,
-    normalize_path_buf, relative_module_path, split_specifier_suffix, strip_extension,
+    RESOLVABLE_SCRIPT_EXTENSIONS, RenameTarget, RenderStyle, apply_all_path_renames,
+    candidate_exists, is_index_file, normalize_path_buf, render_module_specifier,
+    split_specifier_suffix, strip_extension,
 };
 use crate::server::ServerState;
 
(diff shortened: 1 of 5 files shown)


ubugeeei:
@codesmith-bot fix ci

PR #4128, opened 2026-08-10
fix(vite): preserve authored SFC HMR

### LINKED ISSUE #4104: test(vite): exercise authored SFC HMR in a pinned Vite app
## Problem

The repository has one real authored-source HMR oracle: the pinned nuxt-ui dev spec edits Button.vue, observes the Vize virtual module request and Vite hot-update completion, preserves browser state, restores the source, and rejects a page reload.

That is a Nuxt/Vite integration path. There is no equivalent pure Vite pinned-app oracle, so regressions in the direct @vizejs/vite-plugin HMR hook can escape unless they also reproduce through Nuxt.

Closed Issues #2090, #2130, and #2242 fixed stale transformed-module invalidation, while #3403 fixed hot-update lookup cost. This Issue adds durable breadth for those paths; it does not reopen or duplicate their implementations.

## Outcome

Exercise real authored SFC HMR through @vizejs/vite-plugin in a pinned Vite application and make it a deterministic compatibility gate.

## Required work

- [ ] Select one pinned authored Vite fixture with an exact source anchor and stable rendered target.
- [ ] Start the real Vite dev server with Vize and load the authored component in a browser.
- [ ] Edit the authored SFC, observe the transformed .vue.ts?vue&vize request and Vite hot-update completion, and assert DOM updates without reload.
- [ ] Preserve an in-page sentinel to prove component HMR rather than navigation or full reload.
- [ ] Restore the source byte-for-byte on success, assertion failure, worker termination, and supported signals.
- [ ] Cover a direct SFC edit and one dependency/src-owner invalidation path represented by the closed regressions.
- [ ] Reject stale transforms, missing HMR payloads, duplicate updates, fatal console errors, hydration/runtime errors, and page reloads.
- [ ] Add strict inventory and cleanup-contract tests and wire the oracle into the appropriate fast/full compatibility lane.

## Acceptance criteria

- Disabling the Vize HMR hook or cache invalidation makes the oracle fail.
- The test uses a real Vite server and authored source, not a mocked hook.
- The existing nuxt-ui HMR oracle remains intact and independent.
- Relevant App E2E and workflow checks pass.

Child slice of #3227. Refs #2090, #2130, #2242, #3403.

ubugeeei:
## Summary

- exercise direct authored SFC HMR and external template ownership in the pinned Misskey pure-Vite app
- retain dependency-to-owner routing across compiled-value eviction
- restore a known client owner through Vite 8 EnvironmentModuleGraph when the first update removed its graph node
- keep SSR-only owners out of the client graph and fall back to Vite default HMR if restoration fails
- start the repair in the next Vite watcher change window, using the emitted HMR timestamp instead of a retry or blanket delay
- fail closed on reloads, stale transforms, duplicate updates, lost browser state, missing transport events, or incomplete source restoration

## Validation

- old behavior RED: the second external-template update returned an empty module set
- exact Linux artifact isolated Chokidar's 50 ms same-path change coalescing after the forward browser update completed
- latest local Misskey with exact Vite 8.2.0: 8/8 green; direct and external-template update plus repair all emitted HMR without reload
- full Vite plugin suite green
- repository format/lint: 2168/1544 files, zero warnings
- source-length: no new or grown file over 350 lines
- git diff --check

Closes #4104


ubugeeei:
CI root cause follow-up: the first external SFC block update evicted the compiled owner and also removed the dependency-to-owner reverse route, so a repair arriving before Vite reloaded the owner emitted no second update. The fix now separates compiled-value eviction from dependency ownership. Ownership survives until the next compilation reconciles dependencies, while explicit delete and clear still remove it. Latest-main local evidence: full Vite plugin suite green; repository format/lint and source-length green; real pinned Misskey HMR green with two updates for the direct SFC and two updates for the external template. Auto-merge remains disabled until the new head passes CI and exact runtime acceptance.

ubugeeei:
Latest-head Linux HMR failure classification for `5d7e702b039f07fd1d54f2dcc9cfa6cb56fcb333`:

- Job: https://github.com/ubugeeei-prod/vize/actions/runs/31436434321/job/93611575272
- All three attempts completed the external-template forward update: one Vite frame, one 200 transform, and the authored marker was applied.
- The repair write produced no second file event or server HMR update.
- The ineffective `server.watcher.add()` follow-up was removed.
- Vite 8.2's Linux watcher uses Chokidar's 50 ms same-path `change` coalescing. The small external template completed its browser round trip inside that window, so the test's immediate repair was not a distinct watcher event.
- Head `568b0294a` now uses the emitted Vite update timestamp to enter the next change window before the repair. There is no retry, fixed blanket sleep, timeout increase, or weakened transport assertion.
- The production fix—durable dependency ownership and client module restoration—remains unchanged and keeps its old-behavior RED regression.

Auto-merge remains off pending the new exact Misskey lane and all current-head checks.

PR #4233, opened 2026-08-12
fix(patina): ignore unused components when a dynamic `is` binding is present

ubugeeei:
## Problem

`eslint-plugin-vue`'s `vue/no-unused-components` defaults `ignoreWhenBindingPresent` to `true`:

```js
if (node.value.expression.type === "Literal") usedComponents.add(node.value.expression.value);
else if (ignoreWhenBindingPresent) ignoreReporting = true;
```

A non-literal `:is` expression stops the rule for the whole file, because a dynamic `<component :is="resolved">` can render any registered component and the rule cannot tell which. Vize reported anyway.

## Pinned reproduction

`tests/_fixtures/_git/element`, `packages/result/src/index.vue`:

```vue
<component :is="iconElement" :class="iconElement" />
...
components: {
  [IconSuccess.name]: IconSuccess,
  [IconError.name]: IconError,
  ...
}
```

All four icons were reported unused. The registration is under a computed key, so it leaves no statically matchable name — exactly the case upstream declines to guess at.

## Corpus measurement

Via `tools/fixtures/lint-divergence-report.mjs` over all 134 pinned projects:

| | before | after |
| --- | ---: | ---: |
| `vue/no-unused-components` false positives | 2,849 | 2,592 |
| false negatives (all rules) | 23,517 | 23,517 |
| shared (all rules) | 15,089 | 15,089 |

**−257.** I want to flag that I initially estimated ~2,235 from a quick script over the divergence artifacts; that script double-counted repeat findings within a file and the estimate was wrong. The measured number is 257.

## Scope of the suppression

As narrow as upstream's, with a test per boundary:

- a literal `:is="'Panel'"` names its component → still reports
- a static `is="Panel"` attribute is not a binding → still reports
- a dynamic directive argument `:[key]="value"` can resolve to `is` → suppresses
- the pinned `element` shape → clean

## Tradeoff

One true positive is given up: a shadowed import under a dynamic `:is`, which Vize could prove unused (`const current = computed((Child) => Child)`). Its test is **inverted rather than deleted**, with the reasoning recorded in the test itself. Recovering that detection needs the upstream escape hatch (`ignoreWhenBindingPresent: false`), not a divergent default.

## Verification

- `cargo test -p vize_patina` — 2,488 pass, 0 fail.
- `cargo clippy -p vize_patina --all-targets` clean; `rustfmt --edition 2024 --style-edition 2024` clean on both touched files.

Part of #3223.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/4233"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved unused-component detection for dynamic `is` bindings, including nested elements, conditional branches, and loops.
  * Prevented false-positive warnings when component names may resolve dynamically at runtime.
  * Continued reporting genuinely unused components referenced through literal or static `is` attributes.
  * Improved handling of imports, script references, computed registrations, aliases, template literals, and dynamic directive arguments.

* **Tests**
  * Added comprehensive coverage for component usage and ignore behavior.


ubugeeei:
@codesmith-bot fix conflict

Issue #3954, opened 2026-08-09
test(canon): make full-corpus FP/FN parity release-blocking


ubugeeei:
## Problem

The fast compatibility ratchet is valuable, but it is not a proof that a release has no false positives or false negatives across the real-project corpus.

Current evidence:

- The registry contains 134 pinned projects, while only 11 opt into `typecheckPerformance.compareTo = "vue-tsc"`.
- The checked-in fast baseline covers 9 projects; several entries are diagnostic-free `0/0` comparisons.
- `.github/workflows/real-project-matrix.yml` sets `BUDGET_MODE=record-only` for the normal release `workflow_dispatch`. It enforces only on the weekly schedule or when a caller explicitly opts in.
- The workflow comment still says the baseline is unusable via #3513 even though #3513 is closed as completed.

A release can therefore pass while the full-corpus divergence job only records a breach.

## Outcome

Zero unexplained Vize-only and baseline-only diagnostics becomes a release invariant, with proof that both tools actually checked the same authored Vue corpus.

## Re-audit evidence

The Phase 2 source audit found that corpus identity and several classification guarantees already exist and should be preserved rather than rebuilt:

- `materializeBaselineProject` derives vue-tsc's explicit `files` from Vize's sorted checked-file list.
- `evaluateVueProgramCoverage` compares vue-tsc `--listFiles` against that list, records counts and SHA-256 digests, and marks missing or unexpected authored Vue files unusable.
- `evaluateBudget` fails unusable coverage; diagnostic-free `0/0` output is accepted only when both sides checked the same non-empty Vue corpus.
- `compareTypecheckDiagnostics` keeps code/range/message mismatches, false positives, and false negatives distinct.

The remaining production gaps are release enforcement, non-vacuous seeded mutation evidence, corpus breadth, and complete runtime/performance provenance. PR #3959 removes the release-path `record-only` escape hatch.

## Required work

- [ ] Remove the release-path `record-only` escape hatch; a release evidence dispatch must enforce the divergence verdict. PR #3959.
- [x] Require both sides to report the exact same authored Vue input set before computing FP/FN ratios.
- [x] Distinguish a genuinely diagnostic-free project from a vacuous baseline by comparing exact program file sets, file-list hashes, and file counts.
- [ ] Bind source config provenance and dependency/baseline preparation commands to the divergence artifact, then reject mismatched or missing preparation evidence.
[...]

Issue #3767, opened 2026-08-03
vize check: plain objects with a `value` property are wrongly ref-unwrapped in templates under Vue 2.7 compatibility


ubugeeei:
## Summary

With `compiler.compatibility.vueVersion: "2.7"`, Canon's template-scope unwrap helper treats **any object that structurally has a `value` property** as a `Ref`. Plain constant objects such as `{ text, value } as const` — an extremely common shape for select/option definitions — are unwrapped to their `value` type inside the template, so every `OPTION.value` access in a template becomes a false `TS2551`.

This does **not** reproduce on the Vue 3 default path.

## Version

- `vize@0.318.0`
- `vite-plus@0.2.7`
- `typescript@7.0.2`
- Node 26.5.1

## Reproduction

`vize.config.ts`:

```ts
import { defineConfig } from 'vize'

export default defineConfig({
  compiler: {
    compatibility: { vueVersion: '2.7', nuxtVersion: 2, webpackVersion: 4 }
  }
})
```

`PlainValue.vue`:

```vue
<script setup lang="ts">
// Not a ref. Just an object that happens to have a `value` property.
const OPTION = { text: 'Login info', value: 'LOGIN_INFO' } as const
</script>

<template>
  <div>{{ OPTION.value }}</div>
</template>
```

```bash
$ vize check PlainValue.vue
  error:7:18 [TS2551] Property 'value' does not exist on type '"LOGIN_INFO"'. Did you mean 'valueOf'?
  1 error(s)

$ vize check --no-config PlainValue.vue   # Vue 3 defaults
  No type errors found!
```

## Cause

The generated virtual TS declares:

```ts
type __U<T> = T extends import('vue').Ref ? __VizeWidenTemplateRef<T['value']> : T;
```

and then binds every template-scope identifier through it:

```ts
var OPTION: __U<__R_OPTION> = undefined as any;
```

In Vue 3, `Ref<T>` carries a `[RefSymbol]: true` brand, so `T extends Ref` is a nominal-ish check and plain objects do not match. In Vue 2.7 the exported `Ref<T>` is effectively `{ value: T }` with no brand, so `T extends Ref` degenerates into a purely structural "has a `value` property" test and matches any such object.

## Impact

[...]

Issue #2617, opened 2026-07-05
fix(canon): typed template refs do not unwrap exposed component refs


MickaelOth:
### Area

CLI

### Version

0.277.0

### Minimal reproduction

`tsconfig.json`

```json
{
  "compilerOptions": {
    "target": "ES2023",
    "module": "ESNext",
    "moduleResolution": "bundler",
    "strict": true,
    "noEmit": true,
    "jsx": "preserve",
    "jsxImportSource": "vue",
    "lib": ["ES2023", "DOM", "DOM.Iterable"],
    "skipLibCheck": true
  },
  "include": ["src/**/*.vue"]
}
```

`src/Child.vue`

```vue
<script setup lang="ts">
  import { useTemplateRef } from "vue";

  const element = useTemplateRef<HTMLElement>("element");

  defineExpose({ element });
</script>

<template>
  <div ref="element" />
</template>
```

`src/Parent.vue`

```vue
<script setup lang="ts">
  import { useTemplateRef } from "vue";

  import Child from "./Child.vue";

  const child = useTemplateRef<InstanceType<typeof Child>>("child");
</script>

<template>
  <Child ref="child" />
  {{ child?.element?.id }}
</template>
```

Run:

```powershell
vize check --tsconfig tsconfig.json src
```

### Actual behavior

`vize check` reports `TS2339`:

```text
error:11:22 [TS2339] Property 'id' does not exist on type '__ShallowRef<HTMLElement | null>'.
```

The same files are accepted by `vue-tsc` 3.3.5.

This reproduction uses explicit generic arguments on `useTemplateRef`. Removing them makes Vize accept the files, but the typed `useTemplateRef<T>()` form is valid Vue code.

### Expected behavior

`vize check` should pass because refs exposed with `defineExpose()` are available on the parent component instance with Vue's public instance unwrapping. In this reproduction, `child?.element` should be treated as `HTMLElement | null`, so reading `.id` is valid.

### Environment

- OS: Microsoft Windows 10.0.19045
- Architecture: X64
- Node.js: v24.16.0
- Package manager: npm 11.13.0
- Vize: 0.277.0
- Vue: 3.5.39
- vue-tsc: 3.3.5
- TypeScript: 6.0.3

### Before submitting

- [x] I checked existing fix reports and pull requests.
- [x] This is not a private security report.

Issue #913, opened 2026-06-03
fix(typechecker): resolve Nuxt auto-imports and template globals


ubugeeei:
## Summary

`vize check` reports unresolved identifiers for common Nuxt/Nuxt module globals even after the Nuxt type files have been generated. This affects script setup composables and template globals such as `$t`.

## Minimal reproduction

Create a small Nuxt 4 app with `@nuxtjs/i18n`, `@nuxt/content`, and `@vizejs/nuxt` enabled. Use a component like this:

```vue
<script setup lang="ts">
const { locale } = useI18n()
const localePath = useLocalePath()
const page = await queryCollection('content').path('/about').first()
</script>

<template>
  <NuxtLink :to="localePath('/')">{{ $t('nav.home') }}</NuxtLink>
  <pre>{{ locale }} {{ page?.title }}</pre>
</template>
```

Generate Nuxt types, then run Vize check:

```sh
nuxi prepare
vize check app --strict
```

## Actual behavior

Vize reports errors such as:

```text
Cannot find name 'useI18n'.
Cannot find name 'useLocalePath'.
Cannot find name 'queryCollection'.
Cannot find name '$t'.
```

## Expected behavior

For a Nuxt project, `vize check` should resolve Nuxt-generated auto-import declarations and template globals from the generated `.nuxt` type files, or provide a documented way for the Nuxt module to supply those declarations.

## Suggested fix

When the project is a Nuxt project, load the generated Nuxt tsconfig/declaration graph before materializing the Corsa virtual project. In particular, include `.nuxt/imports.d.ts`, `.nuxt/types/*`, and module-generated app declarations used by Nuxt's `tsconfig.app.json`. For template globals, inject or reference the same global component instance declarations Nuxt/Vue tooling uses for `$t` and related helpers.

ubugeeei:
The first trustworthy full parity sweep proves this is still unresolved and release-blocking. Real Project Matrix run 31473431485 at `c00884fc2b40fb60657006ff8dd95a4cfe30008b` ran `pnpm exec nuxt prepare`, bound the clean generated `.nuxt/tsconfig.app.json` (SHA-256 `626ab26d2ecce77a6c87728ab1d182c9dd1c3a223274de089bacefc12e2977a6`), and compared the same 253 authored Elk Vue files.

[...]

PR #4243, opened 2026-08-12
fix(canon): unwrap setup refs consistently across template expressions

### LINKED ISSUE #4146: fix(canon): unwrap setup refs consistently across template expressions
## Severity

P0 — Vize rejects valid Vue 3 templates because setup refs/computed refs are not consistently auto-unwrapped at template use sites. The release parity gate is red across a pinned real application.

## Exact RED

Real Project Matrix run [31473431485](https://github.com/ubugeeei-prod/vize/actions/runs/31473431485) at `c00884fc2b40fb60657006ff8dd95a4cfe30008b` compared the same 253 Elk Vue files with a clean prepared Nuxt configuration.

After separating Nuxt/i18n generated-global diagnostics into #913, 43 false positives remain whose Vize types visibly retain `Ref<T>`, `ComputedRef<T>`, or `WritableComputedRef<T>` where `vue-tsc` applies template auto-unwrapping. Representative failures include:

- `mediaPreviewList.length`: TS2339 because Vize sees `Ref<MediaAttachment[]>`.
- `currentUser.account` / `.server` / `.token`: TS2339 because Vize sees `ComputedRef<UserLogin | undefined>`.
- `instance.configuration`: TS2339 because Vize sees `ComputedRef<ElkInstance | null>`.
- values passed to component props remain `Ref<T>` and cause TS2322/TS2345/TS2740.
- string comparisons remain `string` versus `ComputedRef<string>` and cause TS2367.

These are exact authored diagnostics absent from `vue-tsc`; file coverage and baseline configuration are usable.

## Required contract

- Apply Vue's template unwrapping rules once, at the same semantic boundary used by Vue/`vue-tsc`.
- Preserve optionality and nullability of the unwrapped value.
- Handle bare identifiers, member access, optional chaining, comparisons, component props, `v-if`, `v-for`, `v-model`, and nested expressions consistently.
- Do not unwrap refs in normal script code.
- Do not treat arbitrary objects with a property named `value` as refs, including Vue 2.7 compatibility shapes fixed by #3767.
- Do not double-unwrap values returned from destructured composables or exposed component instances.

## Strict acceptance

- [ ] Add full-equality Vize/`vue-tsc` oracles for `Ref<T>`, `ComputedRef<T>`, `WritableComputedRef<T>`, shallow refs, readonly refs, nullable refs, and union refs.
- [ ] Cover reads, writes, nested member access, optional chaining, array length/indexing, comparisons, component props, `v-model`, and `v-for`.
- [ ] Add negative controls for plain `{ value: T }` objects, script-only reads, Vue 2.7, and explicit `.value` misuse in templates.
- [ ] Assert valid, invalid, and repaired authored diagnostics by complete code/message/range equality; no `any`/`unknown` widening.
- [ ] The exact Elk sweep removes the 43 ref-shaped false positives without changing its Vue program or hiding the remaining separately owned differences.
- [ ] Full Canon/Vize/Content Mapper, Vue 2.7/3 parity, and release matrix gates pass for this surface.

Refs #3954, #3767, #2617.

ubugeeei:
## Summary

`TemplateRefUnwraps::collect` (`crates/vize_canon/src/virtual_ts/generator/template_refs.rs`) built the
template-scope unwrap set **exclusively** from `summary.bindings.bindings` — the SFC's own script
declarations and imports. A binding a framework auto-import injects is not in that map, so
`emit_type_captures` / `emit_template_variables` never emitted its
`type __R_x = typeof x;` + `var x: __U<__R_x> = undefined as any;` shadow pair, and every template use
site kept the declared `Ref<T>` / `ComputedRef<T>` / `WritableComputedRef<T>`. That is the mechanical
source of Elk's remaining ref-shaped false positives.

An auto-imported composable **is** a `<script setup>` import once the framework's unimport transform
runs, so Vue applies exactly the same unwrapping to it as to an authored import. The unwrap set now
includes the auto-import value bindings the template references — at the one boundary that already
owns the rule, not at a new site.

The invariant this establishes is **provenance independence**: the same authored template types
identically whether the composable is imported by the SFC or auto-imported.

### How the candidate set stays safe

`auto_imports::collect` mirrors `emit_auto_import_stubs`' skip rules exactly, so every shadowed name is
guaranteed to have a module-scope declaration in the generated file. That guarantee is load-bearing: a
`typeof` capture of an *undeclared* name would report `TS2304` at a generated position and silently
widen the authored "Cannot find name" diagnostic to `any` — trading a false positive for a false
negative. Declined, with a negative control for each:

| Shape | Why |
| --- | --- |
| `declare function useX(...)` | a function is never a ref |
| `declare const useX: any;` | the degraded no-generated-types fallback; `__U<any>` is `any` |
| `$`-prefixed names | the template context declares them itself (`TS2451` otherwise) |
| `type X = any;` | declares no value |
| component bindings / `external_template_bindings` / `used_components` | already declared in `__template()` |
| configured `template_globals` / `css_modules` | same |
| names the script declares or imports | already own their shadow |
| Vue 2 / 2.7 dialect | its `__U` is the structural `T extends { value: infer V } ? V : T`, which cannot tell a ref from a plain `{ text, value }` constant (#3767) |

The batch virtual project materializes `auto_import_stubs` into one program-wide ambient `.d.ts` and
clears the per-file list, so `VirtualTsOptions` gains `auto_import_bindings` (names only) and
`VirtualTsOptions::auto_import_binding_names()` resolves both spellings identically.

## Change Class

- [x] Virtual TypeScript and type checking

## Behavior Reference

Closes #4146. Refs #3954, #3767, #2617, #913.

Vue's runtime boundary: `PublicInstanceProxyHandlers` applies `proxyRefs`/`unref` to **setupState**
only, never to `globalProperties`. Auto-imports become setupState (the transform rewrites them into
`<script setup>` imports), which is exactly why they are in scope for this fix and arbitrary ambient
globals are not.

## Verification Evidence

All CLI evidence used a privately built copy of the binary (`cargo build -p vize --bin vize` then
`cp target/debug/vize <private path>`), never the shared `target/debug/vize`.

### 1. vue-tsc full-equality oracle + provenance equality (new)

`tests/snapshots/check/template-ref-unwrap-oracle.ts`, registered in `tests/package.json`
(`test:check:fixtures`). Two workspaces built from the **same** authored template — line 3 is the only
difference and both spellings are exactly one line, so every authored line and column is identical:

* `explicit/` — `import { … } from './composables'`, checked by **real `vue-tsc` 3.3.4 / vue 3.6.0-beta.10**
* `auto/` — Nuxt-shaped (`nuxt.config.ts`, `.nuxt/tsconfig.json`, `.nuxt/types/imports.d.ts`), no import

For six variants the test asserts `vue-tsc(explicit) == vize(explicit) == vize(auto)` on the complete,
sorted diagnostic list (code, message, authored line and column):

| Variant | Expected (identical in all three) |
| --- | --- |
| clean | *(empty)* |
| misspelled member on an unwrapped array | `src/App.vue(8,21): error TS2551: Property 'lenght' does not exist on type 'Media[]'. Did you mean 'length'?` |
| explicit `.value` in a template | `src/App.vue(8,21): error TS2551: Property 'value' does not exist on type 'Media[]'. Did you mean 'values'?` |
| plain `{ value }` constant (#3767 control) | ``src/App.vue(17,18): error TS2551: Property 'valeu' does not exist on type '{ readonly text: "Login info"; readonly value: "LOGIN_INFO"; }'. Did you mean 'value'?`` |
| script-only read without `.value` (script control) | `src/App.vue(4,35): error TS2339: Property 'length' does not exist on type 'Ref<Media[], Media[]>'.` |
| repaired | *(empty)* |

The clean template covers `Ref<T>`, `ComputedRef<T>`, `WritableComputedRef<T>`, `ShallowRef<T>`,
`Readonly<Ref<T>>`, nullable `Ref<T | null>`, union `Ref<string | number>` — across reads, writes
(`@click="flag = !flag"`), nested member access, optional chaining, array `length` and indexing,
comparisons, component props, `v-if`, `v-for` and `v-model`.

```
$ VIZE_TEST_BIN=<after> VIZE_TEST_REQUIRE_TSGO=1 node --test snapshots/check/template-ref-unwrap-oracle.ts
✔ template ref unwrapping matches vue-tsc and is provenance-independent (4836.766959ms)
ℹ pass 1  ℹ fail 0
```

RED on the same fixture with a binary built from the reverted source:

```
$ VIZE_TEST_BIN=<before> VIZE_TEST_REQUIRE_TSGO=1 node --test snapshots/check/template-ref-unwrap-oracle.ts
✖ template ref unwrapping matches vue-tsc and is provenance-independent
  AssertionError: an auto-imported composable must type like an authored import for "clean"
  + "src/App.vue(10,24): error TS2339: Property 'account' does not exist on type 'ComputedRef<UserLogin | undefined>'."
  + "src/App.vue(11,24): error TS2339: Property 'token' does not exist on type 'ComputedRef<UserLogin | undefined>'."
  + "src/App.vue(12,14): error TS2367: This comparison appears to be unintentional because the types 'WritableComputedRef<string, string>' and 'string' have no overlap."
  + "src/App.vue(13,23): error TS2339: Property 'length' does not exist on type 'ShallowRef<Media[]>'."
  + "src/App.vue(14,11): error TS2365: Operator '+' cannot be applied to types 'Readonly<Ref<number, number>>' and 'number'."
  + "src/App.vue(15,25): error TS2339: Property 'server' does not exist on type 'Ref<UserLogin | null, UserLogin | null>'."
  + "src/App.vue(20,45): error TS2339: Property 'id' does not exist on type 'true | Media[]'."
  + "src/App.vue(20,57): error TS2339: Property 'url' does not exist on type 'true | Media[]'."
  + "src/App.vue(22,19): error TS2588: Cannot assign to 'flag' because it is a constant."
  + "src/App.vue(23,11): error TS2740: Type 'Ref<Media[], Media[]>' is missing the following properties from type 'Media[]': length, pop, push, concat, and 29 more."
  + "src/App.vue(23,30): error TS2322: Type 'WritableComputedRef<string, string>' is not assignable to type 'string'."
  + "src/App.vue(23,49): error TS2322: Type 'Readonly<Ref<number, number>>' is not assignable to type 'number'."
  - []
```

That RED list is the issue's failure taxonomy in full: `TS2339` on `Ref`/`ComputedRef`/`ShallowRef`,
`TS2367` string-vs-`ComputedRef<string>`, `TS2322`/`TS2740` component props, plus `TS7053` array
indexing, `TS2365` readonly-ref arithmetic and `TS2588` `v-on` writes.

### 2. Rust regression tests (new)

* `crates/vize_canon/tests/auto_import_ref_unwrap.rs` — 8 generated-TS tests: the shadow pair is
  emitted and the capture precedes `__template()`; unreferenced names, function/`any`/`$`/type stubs,
  component names, configured template globals and already-imported names are declined; the
  batch names-only spelling produces a byte-identical `__template()` body; Vue 2.7 keeps its
  structural `__U` for its own setup refs but never takes an auto-import.
* `crates/vize_canon/tests/auto_import_ref_unwrap_parity.rs` — 6 end-to-end tests comparing the
  **complete** diagnostic list (code, message, authored line, column) for the explicit and auto
  provenances, over the same fixtures, with a `vue` stub that keeps the real package's nominal
  `RefSymbol` brand so the plain-`{ value }` control is a real control.

```
$ cargo test -p vize_canon --test auto_import_ref_unwrap --test auto_import_ref_unwrap_parity
test result: ok. 8 passed; 0 failed
test result: ok. 6 passed; 0 failed
```

Mutation check (revert `generator.rs`, `template_refs.rs`, delete `auto_imports.rs`; keep the tests):

```
test result: FAILED. 5 passed; 3 failed     # auto_import_ref_unwrap
test result: FAILED. 1 passed; 5 failed     # auto_import_ref_unwrap_parity
```

Only the new behaviour goes RED — every negative control (`plain_value_objects_…`,
`vue2_structural_unwrap_…`, `declined_stub_shapes_…`, `component_and_template_global_names_…`,
`script_declared_and_imported_names_…`, `auto_import_names_the_template_never_uses_…`) stays green in
both directions, which is what proves they are controls and not a restatement of the fix.

### 3. No false negative: `TS2304` is preserved

A Nuxt-shaped fixture with a genuinely unknown template identifier used twice:

```
after : error:12:11 [TS2304] Cannot find name 'notDeclaredAnywhere'.
        error:12:36 [TS2304] Cannot find name 'notDeclaredAnywhere'.   (2 errors, was 8)
before: same two TS2304 plus the six ref-shaped false positives
```

Both occurrences keep their exact authored positions. Nothing is widened to `any`/`unknown`; no
diagnostic is filtered.

### 4. No churn for projects without generated framework types

`vize check` on three pinned Elk SFCs in the checked-in fixture (no `.nuxt`, so vize falls back to
`declare const X: any;` / generic function stubs) produces **byte-identical JSON** before and after
(`fileCount=3 errorCount=6` both sides, `diff` reports no difference), because those stub shapes are
declined. Non-framework projects have an empty `auto_import_stubs`, so their generated virtual TS is
unchanged — confirmed by the 1051 pre-existing `vize_canon` lib tests (insta snapshots included).

### 5. Gates

```
$ cargo clippy --workspace -- -D warnings -D clippy::wildcard_imports      exit=0
$ cargo fmt --check                                                        exit=0
$ cargo test -p vize_canon                                                 1053 passed; 4 failed
```

The 4 failures are pre-existing and environment-local: they reproduce **identically on this branch's
merge base** with the source reverted (`batch::executor::tests::cli_global_diagnostics_do_not_trigger_session_fallback`,
`…::incremental_fallback::incremental_session_fallback_is_counted_per_check`,
`…::generic_component_listener_payload::…`, `…::directive_anchors::component_event_handler_anchors_at_the_event_name`).
The last two differ only in a TypeScript parameter-name rendering (`'payload' and 'args'` vs
`'payload' and 'payload'`) — a local `tsgo` version difference, not a behaviour change.

Source-file length budget checked against `origin/main`: `generator.rs` **shrinks** 910 → 906,
`template_refs.rs` 117 → 147, `types.rs` 241 → 303, `css_modules.rs` 239 → 243; every new file is
under 350 lines. No over-limit file grew.

- [x] Minimal fixture or focused regression test added or updated
- [x] Snapshot/baseline changes reviewed and explained
- [x] Parity or real-world fixture coverage considered
- [x] Security/audit impact considered
- [x] Performance status or PR benchmark impact considered
- [x] Fuzzing target or crash-reproducer coverage considered
- [x] Editor/LSP scenario coverage considered
- [x] Ecosystem or broad app compatibility impact considered
- [x] Docs, release, or stability notes updated when public behavior changed

## Risk

**Divergences recorded but deliberately not fixed here**

1. **`vue-tsc` cannot type framework auto-imports at all.** On the prepared Elk baseline it reports
   `Property 'currentUser' does not exist on type '{ … }'` at every auto-import use site — it resolves
   template identifiers against the component instance and never reaches the ambient
   `declare global` block. Those are vize *false negatives* and belong to #913 (Nuxt/i18n generated
   globals), not to this fix. That is also why the auto-import side of the oracle is pinned to the
   vue-tsc-checked explicit side rather than to vue-tsc directly; the comparison is transitive and the
   equality is still complete.
2. **Non-framework auto-import providers.** `unplugin-auto-import` writes an `auto-imports.d.ts`
   ambient file that vize resolves through the TypeScript program but never records in
   `auto_import_stubs`, so those bindings are still not unwrapped. Fixing that needs a manifest source
   for the non-Nuxt case; it cannot be inferred from the ambient file alone without shadowing genuine
   globals (which Vue does *not* unwrap) and losing `TS2304`.
3. **Two i18n diagnostics in the Elk ref-shaped bucket** — `AriaAnnouncer.vue(6,20) TS2339 'locales'`
   and `SettingsLanguage.vue(7,21) TS2352` — are about the `useI18n()` return shape rather than
   template unwrapping, and stay owned by #913.

**Not verified locally**

* The Real Project Matrix Elk sweep with a prepared `.nuxt` (needs `pnpm install` + `nuxi prepare` on
  the fixture; disk-constrained here). The mechanism is proven on a Nuxt-shaped fixture built from
  Elk's own composable declarations, which reproduces the exact reported diagnostics before the fix
  and is clean after it.
* `pnpm -C tests test:check` app snapshots (Elk/Misskey/Nuxt UI) — the fixture submodules are not
  checked out in this worktree. The degraded-mode byte-identical result in §4 is the local stand-in.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/4243"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved template ref unwrapping for framework auto-imports.
  - Prevented duplicate name shadowing between auto-imports and script or setup bindings.
  - Improved handling of supported, unused, conflicting, and type-only auto-imports.
  - Preserved correct ref behavior across Vue 2 and Vue 3 templates.

- **Tests**
  - Added comprehensive coverage for auto-import ref unwrapping, diagnostics, and fixture validation.


ubugeeei:
@codesmith-bot fix conflict

PR #4274, opened 2026-08-13
fix(fresco): guarantee terminal restoration across aborts and signals

### LINKED ISSUE #4207: fix(fresco): guarantee terminal restoration across aborts and signals
## Context

Fresco restores terminal modes transactionally and retries cleanup from `Backend::drop`, but the shipped release profile uses `panic = "abort"`, where destructors do not run. Raw-mode `Ctrl+C` is a normal key event for Doctor, while external process signals and aborting panics have separate lifecycle semantics. The remaining lifecycle contract in #4155 therefore needs an explicit, reusable supervisor rather than relying on unwind alone.

## Outcome

Guarantee that every mode owned by an interactive Fresco session is restored after normal completion, returned errors, interactive interrupts, supported external termination signals, unwinding panics, and aborting panics where the platform exposes a safe pre-abort hook.

## Required contracts

- [x] Track raw mode, cursor visibility and shape, bracketed paste, mouse capture, and alternate-screen ownership as one session state machine. (#4217)
- [x] Keep initialization transactional and restoration idempotent; attempt every independent cleanup action after a partial I/O failure. (#4192, #4210)
- [ ] Restore before an aborting panic without depending on `Drop`. (#4219, #4220: Unix complete; Windows pending)
- [x] Handle raw-mode `Ctrl+C` and supported external termination signals without leaving terminal modes active. (#4223, #4224, #4226; Windows-native coverage remains tracked separately below.)
- [x] Chain existing panic and signal handlers without clobbering application ownership. (#4219, #4224)
- [x] Avoid allocation, locks, formatting, and other non-signal-safe work in native signal handlers; defer cleanup safely when required. (#4224)
- [x] Support nested or rejected sessions explicitly instead of silently corrupting process-global terminal state. (#4218)
- [x] Keep injected writers and headless tests independent from process-global stdout. (#4172, #4181, #4218)
- [x] Expose actionable restoration failures without hiding the original application error. (#4210, #4214)
- [x] Add no work to frame layout, paint, diff, or input hot paths. (#4224, #4226: supervision installs before terminal entry; the handler path is independent of frame and input processing.)

## Conformance

- [ ] Injected-writer tests cover normal exit, body error, initialization failure, restoration failure, retry, and unwind.
- [ ] PTY tests assert terminal flags and exact disable/show/leave sequences after `q`, `Esc`, `Ctrl+C`, I/O failure, panic, and supported signals.
- [ ] Unix and Windows platform paths have native lifecycle tests; unsupported cases report an explicit reason. (#4224: Unix and explicit unsupported fallback complete; Windows pending.)
- [x] Tests verify existing process hooks are invoked exactly once and restored or retained according to the documented ownership model. (#4219, #4224)
- [ ] Documentation comments state defaults, process-global constraints, restoration order, abort limitations, and platform support.

## Pull request policy

Keep implementation slices independently reviewable: state machine, panic supervision, signal supervision, and PTY conformance should not be bundled unless their invariants cannot be tested separately. Use conventional English titles and commits. Enable automatic merge only after restoration, compatibility, and performance checks are green.

Related to #4155, #3138, and #3139.

ubugeeei:
## Summary

- add Windows-native raw console mode snapshots so panic and console-control restoration reset the exact pre-Fresco input mode without relying on Drop
- keep Windows output console modes under Crossterm ownership during raw-mode restoration while temporarily enabling virtual-terminal output processing only for emergency presentation resets
- harden Unix and Windows raw-mode snapshot publication with an atomic owner claim and reader-drain protocol, preventing concurrent callers from overwriting the original terminal state
- isolate Windows panic-hook installation coverage in a subprocess so the parent test process never owns the permanent hook
- update Doctor lifecycle docs and focused platform tests for the Windows supervision path and raw snapshot race coverage
- bound the vue-parity fixture lane Rayon fan-out in the composite action and record `rayon_num_threads=4` in the process-budget baseline after repeated runner EAGAIN failures

## Validation

- cargo test -p vize_fresco --lib (241 passed)
- cargo test -p vize_fresco --lib terminal::backend::lifecycle::panic_hook -- --nocapture (5 passed)
- cargo test -p vize_fresco --lib terminal::backend::lifecycle::raw_mode -- --nocapture (4 passed)
- cargo test -p vize --lib commands::doctor::tui -- --nocapture (17 passed, 474 filtered)
- cargo clippy -p vize_fresco --all-targets --all-features -- -D warnings
- RUSTDOCFLAGS=-D warnings cargo doc -p vize_fresco --all-features --no-deps
- cargo fmt --check
- git diff --check
- node --test --test-concurrency=1 tests/tooling/github-workflows-vue-parity.test.ts tests/tooling/github-workflows-check.test.ts (14 passed)
- wc -l confirms `.github/workflows/check.yml` is back to its base 699 lines and split panic-hook test files are below the 350-line source budget

## Platform note

- Attempted cargo check -p vize_fresco --target x86_64-pc-windows-msvc --all-features locally; this macOS environment fails in dependency build scripts before compiling the changed Fresco code because Windows SDK tools are unavailable (windows.h and lib.exe missing). The repository Windows Actions runners should validate the native path.
- Attempted node --test --test-concurrency=1 tests/tooling/source-file-lengths.test.ts locally; this environment fails before source-length assertions because its MoonBit resolver rejects the repository moon.mod `source` key. The line-count regression reported by Actions has been removed and will be revalidated by the GitHub Actions test-scripts job.

Fixes #4207.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal state restoration after panics and control signals on Windows and Unix.
  * Added safer handling for raw-mode transitions, failed updates, and repeated terminal usage.
  * Preserved existing signal and panic-handler behavior while restoring terminal settings reliably.

* **Tests**
  * Added coverage for Windows console handling, panic recovery, signal supervision, and raw-mode cleanup.
  * Strengthened workflow checks for consistent runner resource settings.


ubugeeei:
@coderabbitai review

PR #4388, opened 2026-08-15
chore(davinci): sync origin/main into davinci

ubugeeei:
Periodic catch-up merge of `origin/main` (25 commits, through `666d400b7`) into the davinci integration branch, taken right after the phase-1 exit gate landed.

One conflict, in `crates/vize_patina/src/rules/vue/no_template_target_blank.rs`: main refactored `static_attribute` to return the attribute node and added `attribute_value`/`static_attribute_value` helpers, while davinci had moved node strings into the arena (P1-10). Resolved by taking **main's structure** and adapting the string extraction to davinci's types — `v.content` is already `&'a str`, so main's `.as_str()` is dropped (it is the unstable `str_as_str` feature on this shape, the same trap that needed a hotfix during P1-10) and the two signatures gained the `AttributeNode<'a>` lifetime.

Verified after the merge: `cargo check --workspace --all-targets` clean, `cargo check -p vize_patina` clean, all six workflows still carry `branches: [main, davinci]`, and `tests/tooling/github-workflows-vue-parity.test.ts` passes 2/2. Merged with a merge commit (not squash) so future main syncs keep a shared merge base.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/4388"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789399050&installation_model_id=422833&pr_number=4388&repository=ubugeeei-prod%2Fvize&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fvize%2Fpull%2F4388&signature=38780779fdb78e5f32b4ed031a8d69181df8039c1ce782f64faac2f1d8528e98"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
[...]

PR #4416, opened 2026-08-16
docs(davinci): re-cut phase 2 at the phase-1 exit

ubugeeei:
## Why

`plan/README.md`'s own rule: a provisional task **cannot be picked up for implementation** until the phase is re-cut at its predecessor's exit and expanded to the full contract. Phase 1 has exited, so this unblocks phase 2 — and phase-2.md's own warning said measurements may overrule its guesses. They did.

## What the re-cut changed (not a reformat)

Twelve items, each forced by something phase 1 measured or landed. The load-bearing ones:

- **`ExprRef` must answer the retained-`None` classes — the reason P2-5 splits.** The provisional two-variant enum assumed every expression has an oxc AST. P1-5 measured that `js_ast` is `Some` **only** for one complete TS expression covering the whole text, and P1-9 measured 11.73% of corpus rewrites falling outside it. The escape-variant semantics are now an explicit review point, not a detail.
- **No walk baseline exists.** The provisional said "compare against the P0-3 walk baseline" — but P0-3 recorded _expression re-parse_ counts, and `[traversal]` is an empty reserved section. Combined with P1-13's never-pinned-target miss, this forced the **P2-12 split**: P2-12a owns the phase-start baselines, the walk probe, `[traversal]`, and the pinned target, has **zero deps, and must merge before P2-9** — so phase 2 cannot repeat phase 1's unmeasurable exit line.
- **Drop-freedom is a construction rule with an enforcer**: the arena containers' `needs_drop` const assertion caught two real violations during P1-10, so P2-1/P2-5a state it as a rule and cite that assertion plus grep-zero as acceptance.
- **Alloc counts are the enforced ratchet** — every new bench in the phase lands with its measured count (a seeded `0` now fails); wall stays report-only, and closing P0-4's Blacksmith pending is explicitly _not_ phase 2's job.
- **Three provisional citations were simply wrong and are fixed**: the timing schema is P0-11's `profile-export.schema.json` (not "P0-4"); the transform directory is reached through `vize_atelier_core::steps` with the driver in `src/transform/`; `vize repro` does not exist, so P2-13 adds the command rather than assuming it.
- **wasm32-wasip2 is why size asserts need `#[cfg(target_pointer_width = "64")]`** — the new required lane would break on unguarded ones (exactly the maintainer hotfix P1-10 needed).
- **P2-11's flag is an unfinished deletion with an owner**, not a charter-#26 fallback; the differential lane carries the risk.

Every task now carries `Steps` / `Acceptance` / `Deps` / `Non-goals` (the README requires Non-goals; the provisional blocks omitted them), and the phase ends in a machine-checkable exit gate in the phase-0/phase-1 shape.

## Honest gaps recorded rather than papered over

- **No TS-id covers the Spolvero feed payload** — P2-18 must add the registry entry in its own PR (deliberately not invented here).
- Registry edits owed: TS-22's _From_ column (`P2-12` → `P2-12b`), TS-25's instance list gains P2-9/P2-11/P2-16.
- Does not exist yet, and the plan says so: `crates/vize_disegno`, `crates/vize_davinci_derive`, `walk-baseline.md`, the `tests/fixtures/davinci-matrix/` fixtures, `vize repro`, any wasm32-wasip2 CI lane, and a `[traversal]` reconciliation in `davinci-budgets.test.ts` (which today validates only `[bench]`).

No number appears in this document that has not been run: it states what must be measured and by which command.

## Verification

- `vp fmt --check` clean on both files (oxfmt turned out to be non-idempotent on inline code spans that wrap a line — it strips two spaces of list-continuation indent per pass; seven such spans were rewritten so no code span crosses a newline, and `vp fmt` is now a verified no-op).
- **All 23 referenced TS-ids exist** in `test-suites.md` (verified by grep).
- **39 cited paths exist**; line-number citations verified verbatim (`folio.rs:81` `pub trait Folio`, `lib.rs:13` `#![no_std]`, `span.rs:82`, `relief/core.rs:10-42`, …). To-be-created paths are explicitly marked absent.
- davinci tooling node tests 11/11.

Docs-only: no code touched.

### DIFF
--- a/davinci-road/plan/README.md
+++ b/davinci-road/plan/README.md
@@ -3,7 +3,9 @@
 > [!NOTE]
 > PR-granular decomposition of the [roadmap](../roadmap.md) phases, written
 > for the agent-implements / maintainer-reviews regime (charter #25). One file
-> per phase. This page defines the task format; the phase files are the plan.
+> per phase, plus a sibling `phase-N-tasks.md` once that phase's per-task
+> contracts outgrow the source-length budget (see **File split** below). This
+> page defines the task format; the phase files are the plan.
 
 ## Task format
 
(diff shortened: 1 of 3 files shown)


ubugeeei:
Superseded by #4452, which carries this branch merged in — closing so the davinci queue shows one reviewable head. Nothing is lost: the commits are reachable from `claude/davinci-completion-migration-326a4a` and from this branch, which stays on the remote.

Per #4452's own description: "Supersedes #4388 and #4416 — both are merged into this branch […] Close them in favour of this, or land them first and I will rebase."

PR #4429, opened 2026-08-17
fix(patina): match eslint-plugin-vue on four lint rules

ubugeeei:
## Summary

Real Project Matrix run [31979524200](https://github.com/ubugeeei-prod/vize/actions/runs/31979524200) at `d197098c` reported **974 lint divergence rows** against `eslint-plugin-vue@10.9.2` on shard 9 (`maxFalsePositiveCount 0` / `maxFalseNegativeCount 0` breached on five of six projects). Grouped by rule rather than by project, four rules explain **460** of them, each from a distinct mechanical cause. This PR fixes those four; the rest are measured, root-caused and listed under *Risk*.

| rule | before (FP / FN) | after | cause |
| --- | ---: | ---: | --- |
| `vue/html-quotes` | 92 / 93 | **0 / 0** | reported the inner text, not the value node |
| `vue/prop-name-casing` | 144 / 0 | **0 / 0** | checked the template side, not the declaration |
| `vue/mustache-interpolation-spacing` | 28 / 56 | **0 / 0** | one finding per interpolation instead of one per delimiter |
| `vue/v-slot-style` | 38 / 9 | **0 / 0** | one style for every position instead of three |
| **shard 9 total** | **333 / 641** | **31 / 483** | |

### Root causes

**`vue/html-quotes`** reported `value.loc.start..value.loc.end`, the value text *between* the delimiters, on `PropNode::Attribute` only and never on a component element. Upstream reports `node.value` — the value node, whose range covers the delimiters — for every `VAttribute[value!=null]`, which includes directives, components, and unquoted values (`class=bare`, enclosed by no quote at all). The span now comes from `html_quotes/value_range.rs`, which scans the attribute's own source range for the `=` separator: the first `=` inside an attribute always separates the name — a directive's argument and modifiers cannot contain one — from the value, so `@click='a = b'` still spans `'a = b'`.

**`vue/prop-name-casing`** implemented template attribute casing. Every one of its 144 findings sat at the *identical span* as a `vue/attribute-hyphenation` finding (verified: 144/144 co-located), so it was a second rule name for a rule vize already had, and it matched zero upstream findings. Upstream's rule is about the *declaration*: `onDefinePropsEnter` + `getComponentPropsFromOptions`, default `camelCase`. It now reads `defineProps` through `analysis.macros.props()` / `prop_declaration()` and ports upstream's rejection-shaped `isCamelCase` / `isSnakeCase` verbatim (`prop_name_casing/casing.rs`) — which is why `count` is camelCase and `$attrs` is too (`$` is not in upstream's symbol set).

**`vue/mustache-interpolation-spacing`** emitted one diagnostic spanning `{{ … }}`. Upstream checks the two delimiters independently and reports each on its own two-character token, so `{{text}}` is two findings, not one. Second defect: whitespace detection accepted only `' '` or `'\n'`, so a tab-indented multi-line interpolation (`{{\n\t…\n\t\t}}`) was a false positive — 3 rows on misskey.

**`vue/v-slot-style`** carried a single `Shorthand | Longform` style. Upstream picks per position: `atComponent: 'v-slot'`, `default: 'shorthand'`, `named: 'shorthand'`. That makes `<Comp v-slot="props">` — reka-ui's and nuxt-ui's universal pattern — *correct*, and `<Comp #default="props">` a finding. The 38 false positives were all bare `v-slot` at a component; the 9 false negatives were all `#default` at a component.

## Change Class

- [x] Semantic analysis, lint, and cross-file analysis

## Behavior Reference

`eslint-plugin-vue@10.9.2` — the version `bench/package.json` pins and the divergence harness runs — specifically `dist/rules/{html-quotes,prop-name-casing,mustache-interpolation-spacing,v-slot-style}.js` and `dist/utils/casing.js`. Every expectation in this PR is a full-equality comparison (rule id, severity, file, line, column, endLine, endColumn) produced by `tools/fixtures/lint-divergence-report.mjs`, not a hand-written guess.

### Real corpora, measured before and after with privately built binaries

```
node tools/fixtures/lint-divergence-report.mjs --project <ids> \
  --vize-bin <bin> --budget-mode record-only --output-dir <dir>
```

| project | before (FP / FN) | after (FP / FN) |
| --- | ---: | ---: |
| elk | 7 / 0 | 2 / 0 |
| vue-devtools-v6 | 4 / 2 | 3 / 2 |
| v-charts | 0 / 9 | 0 / 8 |
| vuestic-ui | 235 / 501 | 23 / 428 |
| layoutit-grid | 87 / 129 | 3 / 45 |
| vue-virtual-scroller | 0 / 0 | 0 / 0 |
| **misskey** (shard 10) | 1241 / 141 | **67 / 140** |
| **reka-ui** (shard 4) | 145 / 0 | **7 / 0** |

The before column for shard 9 reproduces the CI artifact numbers exactly, so the local harness is faithful. misskey and reka-ui were measured as corroboration on corpora the failure report never mentioned: **1,773 divergence rows closed in total**, and *no rule other than the four moved in either direction on any of the eight projects*.

misskey also gained a true positive the rule never had: `packages/frontend/src/pages/settings/statusbar.statusbar.vue` declares `_id`, which upstream reports and vize now reports at the same span (misskey `shared` 1113 → 1114).

### Oracle probe (RED-first, full equality)

A scratch project of hand-built probe files driven through the same comparator, so every span is checked against real `eslint-plugin-vue` output:

- pre-fix binary: **9 false positives, 20 false negatives**
- post-fix binary: **31 shared findings, 0 false positives, 0 false negatives**

Covered: single/double/empty/unquoted attribute values; component attributes; `v-if` / `:class` / `@click` / `v-html` directive values; `=` inside a handler expression; `:foo` shorthand; `{{text}}` / `{{text }}` / `{{ text}}` / multi-line; `v-slot`, `#default`, `v-slot:default`, `#header`, `v-slot:footer`, `#[name]`, `v-slot:[name]` at both component and `<template>` positions; `defineProps` in runtime-object, array and type-literal spellings.

### Rust tests, mutation-checked

`crates/vize_patina/src/rules/vue/{html_quotes,mustache_interpolation_spacing,v_slot_style,prop_name_casing}/tests.rs` assert the **complete** diagnostic list per input as `(rule, severity, start, end, covered source text, message)` — no substring matching. Four minimal mutations (report the inner range again; collapse the two mustache findings into one; set `at_component` back to `Shorthand`; stop reading the declaration) fail **15** of the new tests across all four rules; reverting them is green.

```
cargo test -p vize_patina            → 2533 passed; 0 failed
cargo clippy --workspace -- -D warnings -D clippy::wildcard_imports → clean
cargo fmt --check                    → clean
node --test tests/tooling/glyph-corpus-lint-agreement.test.ts
                                     → 2874 files / 8 projects / 0 violations
node --test tests/tooling/cli-lint-contract.test.ts → 7 passed
node --test tests/tooling/{lint-divergence,lint-divergence-ledger,lint-divergence-budget,lint-divergence-report}.test.ts → all passed
```

- [x] Minimal fixture or focused regression test added or updated
- [x] Snapshot/baseline changes reviewed and explained
- [x] Parity or real-world fixture coverage considered
- [x] Security/audit impact considered
- [x] Performance status or PR benchmark impact considered
- [x] Fuzzing target or crash-reproducer coverage considered
- [x] Editor/LSP scenario coverage considered
- [x] Ecosystem or broad app compatibility impact considered
- [x] Docs, release, or stability notes updated when public behavior changed

### Snapshot review

All five PR-gated lint snapshots were regenerated locally and every changed row accounted for. The base binary reproduced each committed snapshot byte-for-byte first, so the diffs are the change and nothing else.

| snapshot | findings | only change |
| --- | ---: | --- |
| elk | 249 → 244 | −2 `prop-name-casing`, −3 `v-slot-style` |
| npmx.dev | 332 → 272 | −59 `prop-name-casing`, −1 `v-slot-style` |
| nuxt-ui | 104 → 35 | −69 `v-slot-style` |
| reka-ui | 384 → 246 | −2 `prop-name-casing`, −136 `v-slot-style` |
| misskey | 3001 → 1828 | −1103 `prop-name-casing`, −67 `v-slot-style`, −3 `mustache…`, **+1** real `prop-name-casing` |

No other rule's count changed in any snapshot; the `html-quotes` rows kept their count and moved by one column on each side. reka-ui's 136 and misskey's 67 `v-slot-style` removals are independently confirmed false positives by the divergence run above.

### One deliberate difference, documented on the rule

A **directive** value that already contains the required quote is left alone — upstream's `avoidEscape` behaviour, applied to directives only. Glyph prints directive expressions with double-quoted JS strings (`:layout="{ type: 'grid' }"` → `:layout='{ type: "grid" }'`) and must then delimit the attribute with single quotes, so reporting it makes `vize fmt` output that `vize lint` rejects: without the carve-out `tests/tooling/glyph-corpus-lint-agreement.test.ts` fails with **1175 introduced findings**, all `vue/html-quotes`. Plain attribute values, which glyph never rewrites, follow upstream exactly (there is a negative-control test for that). The class has **zero rows** on all eight measured corpora. The correct end state is for glyph to keep the attribute delimiter stable; that is a formatter change and belongs in its own PR.

No ledger row was added: `tools/fixtures/lint-divergence.mjs` only lets a documented divergence cancel one false positive against one false negative in the same file, so a difference that produces no rows cannot be expressed there.

### Deferred divergences, measured and root-caused

514 rows remain on shard 9. Every one was inspected against its source line:

| rows | rule | cause |
| ---: | --- | --- |
| 396 FN | `vue/attribute-order` | `is_ordering_barrier` (added by #4123) treats **every** directive as a barrier, so a directive is never reported and never joins the ordering chain. See below. |
| 36 FN / 3 FP | `vue/no-mutating-props` | the FNs are script-side mutations (`area.flex = …` in `<script setup>`); vize only checks `v-model` and inline handlers. The 3 FPs are span mismatches on member-expression mutations. |
| 13 FN | `vue/valid-v-for` | upstream reports a missing key on a custom element under `valid-v-for` as well as `require-v-for-key`; vize reports only the latter. |
| 13 FN | `vue/valid-attribute-name` | `:el$="el$"` — vize accepts `$` in an attribute name, upstream does not. |
| 12 FP | `vue/no-unused-vars` | vize reports an unused *slot* prop; upstream's rule covers template `v-for`/`v-slot` variables but not this shape. |
| 9 FN / 1 FP | `vue/no-template-target-blank` | vize accepts `noopener` **or** `noreferrer`; upstream requires `noopener` **and** (`noreferrer` unless `allowReferrer`). The FP is a `<NuxtLink href="/…">` — a relative href is not an external link to upstream. |
| 8 FP / 5 FN | `vue/require-v-for-key` | span disagreement on the same finding (vize anchors the tag, upstream the element). |
| 7 FP | `vue/no-unused-components` | Vue 2 shells (`shell-dev-vue2`) where the component is used through a render function. |
| 6 FN | `vue/attribute-hyphenation` | `v-model:isSidebarVisible` — vize does not hyphenate-check a `v-model` argument. |
| 3 FN | `vue/no-multi-spaces` | multiple spaces *inside* a binding expression and before `/>`. |
| 1 FN each | `vue/no-v-html`, `vue/require-component-is` | single instances, cause not isolated. |

**`vue/attribute-order` was measured, not guessed.** Removing the directive barrier (one line) was tried and rejected as insufficient: on vuestic-ui it closes 314 of 388 false negatives but opens **26 new false positives**, and it turns previously clean elk into **8 false positives** (elk `attribute-order` 0/0 → 8/0). vize's category table has no `SLOT` category and no handling for `v-bind="obj"` spreads or `v-model` arguments, so it needs upstream's full table, not the barrier removal. Useful for that follow-up: the barrier removal does **not** trip `glyph-corpus-lint-agreement` (0 violations across 2874 files), so #4123's stated justification — that the linter must not reject formatter output — is not actually enforced by that gate, which measures a *delta* and never sees glyph move a directive.

### Other notes

- The `vue/prop-name-casing` rewrite covers `defineProps` only; the Options API `props:` option is not visible to `analysis.macros`, matching the existing gap documented on `vue/no-unused-properties`. Upstream covers both. No corpus row is affected (upstream reported zero Options API violations across all eight projects).
- Public behaviour changed, so the rule tables and prose in `docs/content/**/rules/{all,vue}.md` were updated in all five locales; `docs/content/rules/vue.md` was kept at its original line count for the 350-line budget.
- `crates/vize_patina/src/rule.rs` stays at 435 lines (unchanged); the i18n JSONs grew but sit under an `i18n` path segment, which `tools/moon/cmd/source_file_lengths` excludes. The MoonBit length script could not be executed locally (`moon update` pulls a `moonbitlang/x` revision this toolchain rejects with `Unexpected key 'pkgtype'`) — the budget was verified by hand against `origin/main` line counts instead.

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/4429"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789533980&installation_model_id=422833&pr_number=4429&repository=ubugeeei-prod%2Fvize&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fvize%2Fpull%2F4429&signature=5e4a201ea85a3e5928be94cf74270892ccab1bd81b9fe56fbd530bf59118d20b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

- **New Features**
  - Improved `vue/html-quotes` support for directives, components, quoted and unquoted attributes, with more accurate diagnostics and fixes.
  - Enhanced `vue/mustache-interpolation-spacing` to report opening and closing delimiter spacing separately.
  - Expanded `vue/v-slot-style` with context-specific checks and support for bare `v-slot`.
  - Updated `vue/prop-name-casing` to validate declared props, supporting camelCase and snake_case options.

- **Documentation**
  - Updated localized rule descriptions and guidance to reflect the revised prop-name and slot-style behavior.


ubugeeei:
Follow-up note on the two CI failures from the first run at `cbd1dea0`:

- **vue-parity** was a real hit: `tests/snapshots/check/create-vue-patch-oracle.ts` pinned `vue/html-quotes` at columns 20–38, the inner text. On create-vue's `App.vue` line 6 the value node `"https://vuejs.org/"` opens at column 19 and closes at column 38, so the correct span is 19–39 — the same delimiter-inclusive range `eslint-plugin-vue` reports. Fixed in `f11b07fb`; the suite is green locally (4/4).
- **criterion-ab** failed twice with `The runner has received a shutdown signal` mid-benchmark (`jsx_compile_mode_aware/small_jsx`), before the A/B comparison ran. Infrastructure, not a measured regression — no threshold output was produced either time. The push of `f11b07fb` starts a fresh run.

PR #4452, opened 2026-08-19
feat(davinci): land the phase-2 substrate and drop davinci into the compiler

ubugeeei:
Opens phase 2 of the Davinci program: the phase-start pin, the three substrate
tasks that had no dependencies blocking them, and the change that gives
`vize_davinci` its first production consumer.

**Supersedes #4388 and #4416** — both are merged into this branch, because
P2-12a's measurements had to be taken on a tree that holds main's current
gates, and `origin/davinci` was 64 commits behind. Close them in favour of
this, or land them first and I will rebase.

## What lands

| | |
| --- | --- |
| sync | `origin/davinci` ← `davinci-main-sync` ← `origin/main` (v0.350.0); generated plan matrices re-run |
| **P2-12a** | the pre-S2 traversal baseline and the phase-2 target, pinned before the work that could bias them |
| **P2-1** | `NodeId`, `SideTable`, `Diagnostic` |
| **P2-2** | the pass manager — const-data pipelines, SIL classification, build-time fusion |
| **P2-3** | `PassObserver` and the four in-tree observers |
| **drop-in** | the shipped backends' traversals declared as Davinci `Pipeline`s, tied to the measured walk counts by a law |

Per-task records: [`davinci-road/plan/phase-2-records/`](davinci-road/plan/phase-2-records).

## The measurement P2-12a exists for

P1-13 could not tick "compile bench improvement ≥ target pinned at phase start"
because neither a target nor a phase-start baseline ever existed. This is the
task that stops that repeating.

`vize_atelier_core::walk_probe` counts a **visit** where a stage's descent
dispatches on a template node's kind, and a **walk** at a stage's root entry.
19 sites, not the 4 obvious funnels — and each extra one was found by a number
that read wrong, not by inspection:

- `generate_root_node` handles the single-root case without passing through
  `generate_node`, which reported `stress-wide` DOM codegen as **0 visits**
  while codegen was emitting a 100-attribute element.
- Vapor lowering has **six** child-list walkers; instrumenting only
  `transform_children` understated `stress-deep` by **20×**.

Measured, byte-identical across two runs (`walks / visits`):

| fixture | DOM | SSR | Vapor | transform component |
| --- | ---: | ---: | ---: | ---: |
| small | 2 / 11 | 2 / 16 | 2 / 25 | 8 |
| medium | 2 / 62 | 2 / 118 | 2 / 102 | 33 |
| large | 2 / 86 | 2 / 106 | 2 / 127 | 57 |
| stress-deep | 2 / 134 | 2 / 144 | 2 / 256 | 72 |
| stress-wide | 2 / 3 | 2 / 4 | 2 / 4 | 2 |
| stress-interp | 2 / 1102 | 2 / 2002 | 2 / 3102 | 1001 |

`walks = 2` on every row and an identical transform column across all three
backends: every backend pays the same traversal in full, then walks again.
Vapor lowering re-walks the same children 2.1–2.6×, which is where
region-owning `ui.for`/`ui.if` and fusion aim. `stress-wide` is a 3–4 visit
floor — attribute width is prop work, not traversal work — so it is the target
table's named control row.

## Davinci describes the shipped pipeline

`crates/vize_davinci/src/legacy_plan.rs` declares the three backends' template
traversals as `Pipeline` const data. Every pass is a mandatory barrier because
that is what the pipeline *is* today.

The declaration is tied to reality by a law, not a comment: each backend's
`davinci_walk_baseline.rs` asserts the measured walks equal
`Pipeline::group_count()` on every ladder fixture. This makes P2-11's target
measurable against a declaration rather than prose (`dom_walks_max = 1` is
exactly the claim that `DOM` becomes one group), and makes P2-12b's swap
mechanical.

### A constraint this uncovered, which P2-11 has to answer

The declaration was first written into `vize_atelier_core` as a normal
dependency. **The release gate rejected it, correctly:**

    vize_atelier_core cannot depend on deferred workspace crate vize_davinci

`vize_atelier_core` is published to crates.io; `vize_davinci` is
`publish = false`, as `vize_disegno` will be. This is not a packaging detail —
it is a hard precondition on the strangler plan: **P2-11 cannot move the DOM
backend onto S2 while the Davinci crates are unpublished**, because
`vize_atelier_dom` is published and would need them at runtime, not just in
tests. Publishing them (and accepting semver gating on an IR still in flux),
folding them into an already-published crate, or making the S2 path an
optional-dependency feature are program decisions, and P2-11's contract names
none of them today. Recorded in `phase-2.md`.

The plans therefore live in `vize_davinci` and the backends read them from
their **dev-dependencies** — the shape `davinci_harness` already uses to
instrument published crates. Davinci is exercised against the real pipeline
without entering anyone's release graph.

## Canaries — all proven, none assumed

Six gates were verified by breaking them, per the P0-7 rule:

- a `[traversal]` ceiling edited off its Rust-side pin
- a ceiling with no recorder cell
- a zeroed `[target.phase-2]` value
- a `davinci_plan` that claims one walk where the compiler makes two
- `PassDesc::new` with a mandatory-and-fusable pass → **compile error**
- `Canonical::produce` from an optional pass → **compile error**

The last two are `E0080` with the documented message; the exact rustc output is
recorded in [`p2-2.md`](davinci-road/plan/phase-2-records/p2-2.md), because
the workspace has no compile-fail harness and that gap should be visible rather
than implicit.

## Zero-cost, gated

| bench | allocs |
| --- | ---: |
| `davinci_pipeline_unobserved` | 0 |
| `davinci_pipeline_no_observer` | 0 |

Alloc-identical, both registered in `budgets.toml [bench]`, both reported
`alloc-gated … ok` by `bench-compare`. Observers are statically dispatched, so
the un-observed path has no attachment check at all — stronger than the
contract's "checked once per pipeline". No `dyn PassObserver` is offered, on
purpose.

## Verified locally

- `cargo test --workspace` — green except 8 pre-existing `vize_canon`
  type-checker cases that need `node_modules` (identical with these changes
  stashed)
- `cargo clippy --workspace --all-targets` — clean
- `cargo build -p vize_davinci --target wasm32-wasip2` and
  `cargo check -p vize_atelier_core --target wasm32-unknown-unknown` — green
- `tools/davinci/assertion-lint.mjs` — green, **no allowlist entry added**
  (two methods were renamed at the source instead: `SideTable::contains_id`,
  `Preserved::preserves`)
- budgets + traversal reconciliation suites — 9 tests green

## Remaining CI failure — inherited from the main sync

`test-scripts` (and `test-report`, which aggregates it) fails on one line:

    | 745 | 743 | over-limit file grew | editors/vscode/test/suite/extension-host.cjs |

Those two lines are `signatureHelp: false,` and `"signatureHelp.enable",`, added
by main's #4444 and pulled in by the merge. Nothing in the davinci commits
touches that file, and shrinking someone else's test harness by two lines to
satisfy a line-count gate would be gaming it rather than honouring it — the gate
exists to force splitting, not cosmetic deletion.

This is structural, not incidental: the gate compares against the PR base, so
every main → davinci sync re-imports main's own growth as new. **#4388 fails
the same way**, and has since 2026-08-15.

The clean fix needs no code: land #4388 into `davinci` first, and this PR's
base then already contains main's growth, so the file has not grown relative to
it. That also resolves the supersedes-#4388 overlap. Alternatives are to waive
the gate for sync PRs, or to compare sync PRs against `main`. All three are
maintainer calls.

Everything attributable to these commits is green: `fmt-rust`, `check-js`,
`publish_crates`, and the source-length gate for all six files the davinci work
touches — the probe was reworked to count inside the iterator constructor so it
adds no source line at all.

## Not met, stated rather than waived

P2-12a's acceptance clause "corpus-coverage `--check` green with scope proof"
**is not evaluable by anyone today**: it byte-compares against a hydrated
corpus that neither CI nor a working tree has, which
`corpus-coverage.md`'s own header already records. The audit's substance was
done against the committed 142/142 report — every S2 op has real-project
instances except `mathml`, which has none at all and is now recorded as "not
represented — matrix fixtures only". The residual clause carries to the exit
gate's C-14 line. Written up as a plan finding in
[`p2-12a.md`](davinci-road/plan/phase-2-records/p2-12a.md).

The `[target.phase-2]` **numbers** are a maintainer review point. The values
here are derived from the measurement above; CI owns only their existence and
non-zero-ness, because the assurance doctrine forbids choosing them later to
fit the result.


ubugeeei:
Diagnosis of the current `test-scripts` failure, from the session that was driving phases 0–1 (my #4388 / #4416 were closed into this PR).

**The MoonBit breakage is gone.** Earlier runs on #4388/#4416 failed inside a third-party dependency (`moonbitlang/x@0.4.47`, `path/win32/path.mbt`, `Invalid lexscan target … got StringView`) because the compiler floats — `install-moonbit.mjs` pins the installer's SHA256, not the toolchain. This PR's log has zero `lexscan` hits, so that resolved itself upstream. Worth pinning the toolchain anyway, since it will float again.

**What actually fails now is the 350-line source budget, and most of it is inherited from main, not authored here.** `tests/tooling/source-file-lengths.test.ts` runs the checker with `--limit 5`, so the report truncates; the full violation set against `origin/davinci` is 15 files:

| file | davinci | this PR | main |
| --- | ---: | ---: | ---: |
| `tools/fixtures/tool-matrix-run.mjs` | 334 | 532 | **532** |
| `tests/_fixtures/npmx-e2e-registry-fixtures.ts` | 0 | 432 | **432** |
| `npm/cli/schemas/vize.config.schema.json` | 971 | 979 | — |
| `npm/native/index.d.ts` | 1146 | 1150 | **1150** |
| `crates/vize_carton/src/i18n/{en,ja,zh}.json` | 452 | 457 | **457** |
| `docs/content/{fr,ja,pt-BR,zh-CN}/rules/vue.md` | 483 | 484–485 | — |
| `editors/vscode/test/suite/extension-host.cjs` | 743 | 745 | — |
| `npm/cli/src/types/generated.ts` | 654 | 658 | — |
| `tests/_fixtures/vue-ecosystem-fixtures.json` | 3515 | 3517 | — |
| `tests/snapshots/check/create-vue-patch-oracle.ts` | 730 | 733 | — |

Spot-checked against `origin/main`: the sizes in this PR match main's exactly for the files in the last column — i.e. no davinci commit touched them; they grew on main and the merge inherits the growth.

**This is structural, not a one-off.** The rule is "an over-limit file that grew relative to the PR base fails", and for a long-lived integration branch the base is `davinci`, so every legitimate main-side growth of an already-over-limit file surfaces as a violation the moment main is merged in. My earlier sync (#4388) passed this check only because main had not yet grown those files; at 65 commits it is unavoidable, and it will recur on every future sync.

Three ways out, in the order I would consider them:

1. **Teach the checker about merges.** For a commit with two parents, fail only when the file grew beyond *both* parents — that is exactly "this branch added lines", which is what the budget is trying to ratchet. Fixes every future sync, but it edits `tools/moon/cmd/source_file_lengths/main.mbt` and therefore affects main too.
2. **Split the two files that genuinely grew a lot** (`tool-matrix-run.mjs` 334→532, and the new 432-line `npmx-e2e-registry-fixtures.ts`) **on main first**, then sync. Principled, but it is main-side work and the small +2/+4/+5 growths would still trip.
3. **Exempt the inherited set for this PR only.** Unblocks now, recurs next sync.

Caveat on verifying any of these locally: the MoonBit runtime on the shared dev machine rejects this repo's `moon.mod` layout, so `source-file-lengths.test.ts` cannot run there — which is precisely why this class of failure keeps reaching CI. This shell equivalent reproduces the CI rule exactly and needs no moon:

```sh
git diff --name-only origin/davinci..HEAD | grep -E '\.(rs|ts|tsx|js|jsx|mjs|cjs|mts|cts|vue|css|scss|html|pkl|mbt|scm|lua|json|toml|yaml|yml|md)$' | while read -r f; do
  cur=$(git show "HEAD:$f" 2>/dev/null | wc -l | tr -d ' '); [ -z "$cur" ] && continue; [ "$cur" -le 350 ] && continue
  b=$(git show "origin/davinci:$f" 2>/dev/null | wc -l | tr -d ' '); b=${b:-0}
  [ "$cur" -gt "$b" ] && echo "VIOLATION $f: $b -> $cur"
done
```

Issue #4461, opened 2026-08-19
fix(matrix): restore the typecheck-divergence release gate


ubugeeei:
`typecheck_divergence_mode` in the release dispatch is temporarily `record-only` so the shipping backlog can move. This issue tracks restoring it to `enforce`.

## Why it was lowered

Real Project Matrix's `typecheck-divergence` surface is the **only** surface the release preflight enforces (`core_tools`, `typecheck_dependencies`, `lint_divergence` and `lsp` are all dispatched `record-only`). It has not been green since **2026-08-12 `af40a9981`**, which blocked v0.348.0, v0.349.0 and v0.350.0 — three tags, three preflight failures, three automatic tag rollbacks, and nothing published since **v0.347.7 on 2026-08-11**.

## What actually breaches — all three are instrument defects

| fixture | verdict | cause |
| --- | --- | --- |
| primevue | 204 false negatives, ratio 0.155 vs 0.05 | the generated baseline extends `packages/primevue/tsconfig.json` (`target: es5`, no `@/volt/*` paths) but `files`-lists `apps/volt/**` and `apps/showcase/**`, which are Nuxt apps whose real config is `apps/<app>/.nuxt/tsconfig.json`. vue-tsc emits TS2307×99, TS7006×64, TS2304×13, TS2802×7; Vize resolves each file under its own nearest tsconfig and does not. **Vize is right; the baseline is wrong.** All 204 are in `apps/volt`. |
| reka-ui | baseline unusable | two `@vue/runtime-core` identities — the foreign one is Vize's `3.6.0-beta.10` |
| elk | baseline unusable | two `@vue/runtime-dom` identities, same shape |

Regression window: `b3b91473c` (#4235) moved the workspace to Vue `3.6.0-beta.10`; `5b1eb281d` (#4236) moved the primevue pin `0394f1e41` → `d4374cb7c`. Both landed hours after the last green run. `e632bf09d` (#4430) then added `typecheck-baseline-ambient.mjs`, which converted the reka-ui/elk mis-measurement from silent to loud — correct behaviour, not a regression.

## Work to restore the gate

**primevue** — approach decided: measure each app under its own tsconfig rather than narrowing coverage. Everything it needs is confirmed:
- nothing enforces unique `fixturePath` (134 entries), so per-tsconfig entries over one path need no harness change
[...]

PR #4464, opened 2026-08-19
perf(release): reuse the parent's gate evidence for a version-only release

ubugeeei:
## The problem

A release commit rewrites the same version into 27 tracked files and regenerates two lockfiles. It touches **no source**. v0.350.1:

```
26 files changed, 88 insertions(+), 88 deletions(-)
```

The preflight re-dispatched **every** gate at the tag SHA anyway and waited for all of them. Measured wall-clock across recent runs:

| gate | minutes |
| --- | --- |
| **Real Project Matrix** | **123, 125, 126, 303, 317** |
| Native Smoke | 12–40 |
| Fuzz | 3–33 |
| Check | 8–11 |
| Benchmark | 4–5 |

The whole release wait is one workflow re-proving code that did not change. Real Project Matrix runs 22 shards — each one installs a real project, builds the CLI, and runs a vue-tsc baseline over thousands of Vue files — against a 20-job hosted cap, so it always takes two waves. For a version string.

## The change

When the release commit's diff against its first parent is version metadata only, the gates that prove the **code** accept the parent's green run and are **not dispatched at all**:

`Check`, `Benchmark`, `Fuzz`, `Miri`, `App E2E`, `Real Project Matrix`, `Docs build`

`Native Smoke` deliberately stays pinned to the tag SHA — it installs what the tag builds, so it is the one gate whose subject really *is* the release commit. The Release workflow's own build and `Smoke release npm package installs` jobs were already at the tag and are untouched.

Result: the release verify drops from **hours to minutes**, because for a version-only tag there is nothing left to wait for.

## Why the path allowlist is safe here

It rides on checks that already existed, all of which must pass first:

- the commit is the **single-parent** release commit (`releaseParentSha`)
- it is on main's **first-parent chain** (`assertReleaseCommitIsOnMainFirstParent`)
- the **remote tag points at it** (`verifyGitReleaseTarget`)
- `assertReleaseMetadata` agrees the version it carries is the tag's

A hand-written commit touching `Cargo.toml` cannot reach the reuse without also faking all of those. And a diff that cannot be computed — shallow clone, unhydrated parent — answers "no" and pays for the full dispatch, so every uncertain case falls back to proving it.

## Verification

`node --test tests/tooling/release-preflight*.test.ts` → **62/62 pass** (58 existing + 4 new).

The new tests assert, against the real gate wiring:

- the actual v0.350.1 file list is recognised; adding any one of `crates/**/*.rs`, `editors/**/*.cjs`, `tools/github/*.mjs` or `.github/workflows/*.yml` defeats it
- only the seven code gates get the parent SHA; `Native Smoke` does not
- a version-only release **dispatches nothing** and selects the parent-SHA run for Real Project Matrix while `Native Smoke` resolves to the tag SHA
- without the reuse, that same parent evidence does **not** satisfy the gates and the matrix is dispatched

`oxlint` and `oxfmt --check` clean; no file crosses or grows past the 350-line budget.

Refs #4461

---
<a href="https://app.blacksmith.sh/ubugeeei-prod/codesmith/vize/pr/4464"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789731645&installation_model_id=422833&pr_number=4464&repository=ubugeeei-prod%2Fvize&return_to=https%3A%2F%2Fgithub.com%2Fubugeeei-prod%2Fvize%2Fpull%2F4464&signature=3bd3300ffe53f90b16161817a9bff95823fab2dc3025f395f4db915ee47eb5ad"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

## Summary by CodeRabbit

* **New Features**
  * Release preflight can reuse qualifying workflow evidence from the release’s parent commit for version-only changes.
  * Native Smoke validation remains tied to the release commit.
  * Unnecessary workflow dispatches and waits are avoided when valid evidence already exists.

* **Bug Fixes**
  * Release checks now handle alternate evidence commits and report all accepted evidence sources when validation is missing.
  * Evidence reuse is disabled safely when change detection cannot be completed.

### DIFF
--- a/tests/tooling/release-preflight-parent-evidence.test.ts
+++ b/tests/tooling/release-preflight-parent-evidence.test.ts
@@ -1,0 +1,135 @@
+import assert from "node:assert/strict";
+import { test } from "node:test";
+
+import {
+  bootstrapRequiredWorkflowRuns,
+  createReleaseGateDispatchPlans,
+} from "../../tools/github/release-preflight-bootstrap.mjs";
+import { isVersionMetadataOnlyRelease } from "../../tools/github/release-preflight-core.mjs";
+import { releaseEvidenceShas } from "../../tools/github/release-preflight.mjs";
+
+const tagSha = "a".repeat(40);
+const parentSha = "b".repeat(40);
+
+/**
+ * The release commit the script writes: the same version into 27 tracked files
+ * plus the two regenerated lockfiles, and nothing else.
+ */
+const versionOnlyPaths = [
+  "Cargo.lock",
+  "Cargo.toml",
+  "README.md",
+  "editors/vscode/package.json",
+  "editors/zed/extension.toml",
+  "npm/cli/package.json",
+  "pnpm-lock.yaml",
+  "pnpm-workspace.yaml",
+];
+
+test("version metadata paths are recognised and source paths are not", () => {
+  assert.equal(isVersionMetadataOnlyRelease(versionOnlyPaths), true);
+  assert.equal(isVersionMetadataOnlyRelease([]), false);
+  for (const source of [
+    "crates/vize_canon/src/lib.rs",
+    "editors/vscode/test/suite/extension-host.cjs",
+    "tools/github/release-preflight.mjs",
+    ".github/workflows/release.yml",
+  ]) {
+    assert.equal(
+      isVersionMetadataOnlyRelease([...versionOnlyPaths, source]),
+      false,
+      `${source} must defeat the reuse`,
+    );
+  }
+});
+
+test("only the code gates reuse the parent; Native Smoke stays pinned to the tag", () => {
+  const shas = releaseEvidenceShas({ sha: tagSha, baseSha: parentSha, versionOnly: true });
+  assert.deepEqual(shas.get("Real Project Matrix"), [tagSha, parentSha]);
+  assert.deepEqual(shas.get("Check"), [tagSha, parentSha]);
+  assert.equal(shas.get("Native Smoke"), undefined);
+  assert.equal(
+    releaseEvidenceShas({ sha: tagSha, baseSha: parentSha, versionOnly: false }).size,
+    0,
+  );
+});
+
+function greenRun(
+  workflowPath: string,
+  headSha: string,
+  displayTitle: string,
+  event = "workflow_dispatch",
+) {
+  return {
+    conclusion: "success",
+    created_at: "2026-08-19T00:00:00Z",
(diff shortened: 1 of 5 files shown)


ubugeeei:
Superseded by #4465 — closing rather than merging, because after that change this buys nothing on the critical path and costs a conflict.

**Why it no longer helps.** #4465 removes `Real Project Matrix`, `Native Smoke` and `App E2E` from the required set. What is left is `Check`, `Miri` and `Docs build` — all push-triggered, so the release commit's own push to main produces them — plus `Benchmark` (4–5 min) and `Fuzz` replay (3 min).

This PR would let a version-only release take `Check`/`Miri`/`Docs build` from the parent instead of waiting ~11 minutes for the release commit's own runs. But the Release workflow's build matrix is ~17 minutes regardless (Windows native: 17 / 14 / 14 / 13), so those 11 minutes were already hidden underneath it. Saving them changes the release wall-clock by zero.

That leaves added machinery — a path allowlist, an `evidenceShas` map threaded through three modules — on the release gate, at a moment when we are deliberately weakening that gate and every extra moving part is risk.

**Worth reviving if the gates come back.** If #4461 restores `Real Project Matrix` / `App E2E` / `Native Smoke` to the required set, the reuse becomes valuable again immediately — a version bump would no longer have to re-run hours of matrix. The branch stays on the remote, and the reasoning plus the four tests are here to lift.

ubugeeei:
Correction: this had already merged when I posted the note above — my close attempt failed against an already-merged PR, so read that comment as context, not as a decision.

Keeping it. #4465 still removes the three hour-long gates, and this reuse now sits underneath that: harmless while `Check`/`Miri`/`Docs build` are the only push-triggered gates left, and immediately valuable again if #4461 restores `Real Project Matrix` / `App E2E` / `Native Smoke`. #4465 will be rebased onto this.

Issue #6258, opened 2026-09-19
fix(content-mapper): preserve disjoint diagnostic directive ownership


ubugeeei:
## Problem

The current TypeScript Content Mapper diagnostic directive protocol accepts one contiguous generated interval per directive. Vize projects multiple authored-node spans into their bounding interval. Generated checks for a child can occur between checks for the parent, so a parent `@vue-expect-error` incorrectly hides the child diagnostic. Splitting an expectation into multiple tuples is also incorrect: the current upstream implementation tracks usage independently for each tuple and reports unused expectations even when another interval of the same authored comment suppressed an error.

Reference: [TypeScript applyContentMapperDiagnosticDirectives](https://github.com/microsoft/TypeScript/blob/d6c4afddb2c55f4a9dea7b59293a99a8fdea1799/tsc/internal/compiler/program.go#L1493). Reproduced while adopting the full Vue Language Tools suite in #6253, under #6239.

```vue
<script setup lang="ts">import Comp from './Comp.vue';</script>
<template>
  <!-- @vue-expect-error -->
  <Comp :id="outer">{{ child }}</Comp>
</template>
```

The expectation may suppress the parent's `outer` error; `child` must remain visible. Test both native and component parents, nested children, multiline props, and edits that repair the parent only. Native Canon/CLI/Maestro ownership is being fixed independently through authored AST ranges; this issue covers the separate standard Content Mapper protocol.

## Acceptance

- [ ] Represent disjoint generated ranges without swallowing intervening child diagnostics.
- [ ] Aggregate used/unused status once per authored expectation across all its generated intervals; do not duplicate unused diagnostics or use broad bounding ranges.
- [ ] Confirm the pinned TypeScript backend understands the representation, with protocol and real tsgo CLI/LSP tests.
- [ ] Match native Canon and Maestro behavior for ignore, expect, skip, orphan and repaired expectations, with exact authored ranges.
- [ ] Verify terminal Actions checks before closing.

Issue #6239, opened 2026-09-19
chore(release): track v1 alpha quality for Vue Fes Japan 2026


ubugeeei:
## Goal

Ship a reviewable `1.0.0-alpha.N` for the public Vize introduction at Vue Fes Japan 2026. Correctness, stability, performance, extensibility, compatibility, and day-to-day editor polish all count as release quality. A green unit suite or an implemented capability is not sufficient proof.

## Scope and release decision

Prioritize the compiler, linter, Canon/typechecker, Maestro/LSP, shipped editor extensions, fresh installation, documentation, and publication. Keep UI/Compose/Framework/Native expansion in its existing roadmap issues; the alpha must accurately label their support level. Do not silently close or claim those roadmaps complete. The exact version and publish decision follow the go/no-go evidence below.

Reuse milestone **v1 alpha production readiness** and existing issues. Its old completed count is historical; this tracker is the current release checklist. The detailed typed-tooling contract remains #3957.

## Delivery batches

Related defects should share a PR when they exercise the same subsystem and validation matrix. Prefer a coherent multi-issue batch over one PR per assertion. Each PR needs conventional naming, attributable fixes, exact regression evidence, and terminal Actions results. Use Actions for broad/host validation and small local reproductions for rapid RED/GREEN feedback. Do useful independent investigation while CI runs; do not rerun unchanged successful jobs without a reason.

- [ ] **Project semantics and safe edits:** #3984, #3953, #6230, #6235, #6236, #6238, #6240, #6242, #6247; preserve lexical symbol identity, authored mappings, declaration output, unsaved overlays, and mixed Vue/TS/JS project behavior.
- [ ] **Real editor workflows and robustness:** #3952, #6233; packaged editor behavior, config/dependency/file lifecycle, cancellation, restarts, stale-result suppression, and host setup.
- [ ] **Compiler/linter quality and compatibility:** #6100, #6097, #6176, #6190, #6186, #6234, #6241; investigate crashes and baseline disagreements instead of blessing existing snapshots.
- [ ] **Performance and efficient evidence:** #6189, #6185, #6243; cold/warm/edit/large-workspace latency and memory, reproducible pinned comparison, complete real-app CI.
- [ ] **Fresh setup, documentation, publication:** #3956, #6101, #6086; packed artifacts on supported hosts, actionable errors, tested instructions, exact-version registry visibility.
[...]

Issue #6176, opened 2026-09-16
fix(tooling): align patterned template compiler and Canon with RFC 823


ubugeeei:
## Contract

Implement the current long-form design documented and tested by https://github.com/vuejs/rfcs/pull/823#issuecomment-5692635473. References: core #15531 checked head 83c5fcc2f27b1e47b9c11dfb01e542fa24c10710; language-tools #6207 checked head 1e89b84018bba8b9609315fbc49cce89ecdca9d0. These are Draft implementations, not a finalized Vue specification.

## Confirmed Gaps

The compiler currently performs permissive string splitting and lowers matching through ordinary single-item v-for loops. Generated names are fixed. Guards and render bindings independently destructure the subject, causing repeated reads and distinct rest values. `as` requires the obsolete `as const name` form. The compiler grammar/lowering is still not converged with the strict shared pattern model. Croquis AST/scopes shipped in #6188; Canon single-SFC/project/CLI/WASM/Playground typing shipped in #6192. Native Maestro opt-in, authored diagnostics, unsaved repair and root-pattern binding hover/definition shipped in #6196. Content Mapper routing and the full navigation matrix remain open.

## Acceptance

- [x] #6175: object/array shapes and real-container preservation, with real VDOM/Vapor/SSR regressions.
- [ ] Shared parsed pattern representation with source spans: strict binding/value/literal/or/as/object/tuple/rest/guard grammar, duplicate and structural diagnostics.
- [ ] Single subject/value/property evaluation, shape checks before reads/copies, identical guard/render bindings, correct object and array rest semantics.
- [ ] Hygienic lexical lowering across nested matches, setup/props, loops, slots, events; preserve authored host/key/v-once and test SSR hydration.
- [x] Canon opt-in propagated across project, single-SFC, CLI and Playground paths. No silent success for unchecked patterns.
- [x] Checked declaration-file type algorithm (not per-template string source), narrowing and subtraction for imported generics, optional keys, tuple combinations, readonly/array rest, conservative any/unknown/open unions, guarded-arm coverage.
- [x] Authored diagnostics for missing coverage (CLI nonzero), unreachable arms (warning-only CLI zero), duplicate declarations and malformed syntax.
- [ ] Maestro definition/rename/completion/diagnostic boundaries for match subject, pattern bindings, guards and nested closures, including `v-for="rows in rows"`.
[...]

Issue #6094, opened 2026-09-13
feat(compose): add SSR data-fetch package


ubugeeei:
## Parent

Part of #3136.

## Outcome

Ship a data-fetch/resource package with typed request/response/error metadata, SSR snapshots, hydration transfer, caching, invalidation, streaming, generator metadata, and source-owned `.vue` usage fixtures.

## Scope

- Idle, pending, success, stale, cancelled, unsupported, and error states.
- Request deduplication, race policies, abort reasons, retries, backoff, deadlines, optimistic work, and stale-while-refresh.
- Server rendering, serialized payloads, client revalidation, and leak/race assertions.
- Fetch adapters, stream adapters, and test adapters.

## Acceptance

- Type inference preserves endpoint literals and payload shapes.
- SSR and hydration fixtures prove no duplicate network work by default.
- Cancellation and stale-response tests are deterministic.
- Public docs/examples use `.vue` SFCs.

PR #6260, opened 2026-09-19
fix(tooling)!: enforce upstream diagnostic and dependency contracts

### LINKED ISSUE #6253: test(canon): adopt upstream Vue Language Tools diagnostic suites
## Problem

Adopt Vue Language Tools' upstream TypeScript error/coverage suite as an independent Canon/LSP oracle, as requested during #6239. Existing Vize authored error fixtures and the Maestro scorecard do not import or execute the full upstream `test-workspace/tsc` suite.

Pinned upstream: vuejs/language-tools@88e8500c1e5f1b29d80f42c8ca065cc9cbd56899 (MIT). Inventory: 230 fixture projects, 599 original fixture/configuration files (including 365 source files), 71 `@ts-expect-error` / `@vue-expect-error` annotations. Preserve original fixtures and expected diagnostics, license and revision provenance.

Patterned-template references are pinned separately at language-tools@e68d41bfef1770f8428f92097abe959b57de62d2 (PR vuejs/language-tools#6207) and core@53455195068f6916c9864a2490e017b584900874 (PR vuejs/core#15531), under #6176.

## Acceptance

- [x] Vendor or fetch the pinned complete source corpus with a reproducible inventory and MIT notice; do not silently omit failing projects.
- [ ] Execute every project's authored tsconfig semantics through Vize, preserve project isolation, and compare expected diagnostic code/severity/authored range and absence of unexpected diagnostics.
- [ ] Enforce error directives, including unused-directive failures; introduce deliberate mutations to show that suppressed/missing diagnostics fail the suite.
- [ ] Exercise repairs through the real LSP to ensure current-buffer diagnostics disappear/reappear at correct document versions.
- [ ] Cover the entire patterned-template typechecking reference (40 coverage cases in nested/root/root-lang-html forms, plus narrowing/generic/rest/event/scope/duplicate fixtures) and language-core helper type assertions.
- [ ] Record every compatibility discrepancy with a concrete reproducer and resolve the common projection/scope/configuration cause. A report or allowlist alone does not complete this issue.
- [ ] Run the complete suite in Actions with bounded concurrency, concise failure summaries and artifacts; link it from the release gate.

## Current evidence — PR #6260, head `9002bfc11`

The strict runner executes all 230 projects with 599 byte-identical original files, 39 expected diagnostics and 71 suppression annotations. **206 pass, 24 fail**. No failing project is excluded or assigned a passing baseline. Functional/dynamic slot inference resolves upstream #5474; declared, indexed and generic slot payloads are tested through native CLI, live LSP and independent vue-tsc valid/error/repair cases. Published generic Vue declarations retain setup-context types, including the real vue-virtual-scroller dependency.

The shared contract covers canonical dependency identity, AST-owned error directives, macro ownership, inherited/top-level compiler options, failed-resolution invalidation, live project membership, workspace synchronization and shared CLI/editor JSX projections. Native model overloads follow the installed Vue version; defaults retain public props/events and generics. Authored diagnostic positions require explicit mappings, with one owner per authored condition. Public generic callable, emit, expose, inferred/declared slot, root-element and recursive-component contracts retain contextual types. Runtime builtins, directive hooks, reserved Options API bindings, CSS Modules and Art dependencies share these contracts.

The latest own-slot work aligns `useSlots()` and `$slots` with declared/inferred payloads under `inferTemplateDollarSlots`, including generic parameters and inherited/top-level options. Nine native/reference/authoring cases, 1,419 Canon tests, 379 Croquis tests and 3 handler regressions pass. Alias/namespace helpers, exact Markdown/hover/definition behavior and incomplete calls retain authored ownership. The `slots` corpus project now has only its four unresolved `Self` payload assertions; no fixture or oracle was changed to hide them.

`inferTemplateDollarAttrs` now aligns `useAttrs()` and `$attrs` with augmented component attributes and native fallthrough types, resolving upstream `#5106` and `attrs`. Seven native CLI/LSP/reference and rich-authoring cases cover error/repair, inherited/top-level options, aliases, namespace imports, Markdown examples and exact navigation after incomplete edits. Hook and variable hovers reject private generated aliases. Related slot/CSS contracts, 1,419 Canon tests, 379 Croquis tests and 14 policy checks pass.

Retained TypeScript scripts now use native syntax diagnostics, including an explicitly mapped block EOF so an unfinished expression cannot consume the next block or generated helpers. Recoverable ambient declarations retain module ownership without synthetic TS1184 errors. All 32 ordinary/setup, LF/CRLF and Unicode invalid/repair states agree across native CLI, live LSP and independent vue-tsc. Native collection preserves parser fallback when the backend is unavailable. Canon's 1,419 unit tests and Maestro's 105 diagnostic tests pass; shared integration fixtures now resolve pinned real Vue types.

The local packaged VS Code host passes all scenarios, including attribute Markdown/signature/example/definition checks after incomplete edits. The host harness now resolves candidates beyond the first item and canonicalizes the external definition fixture path on macOS. Current-head Actions remain pending.

Memory issue #6264 is repaired and verified in Linux Actions. The unchanged full local churn gate passes at 293.6 MiB process-tree peak (384 MiB limit), at most three processes, 40 edits, 290 publications and 20 file lifecycle iterations. No GC tuning or periodic restart is used. LSP executor wake loss is tracked by #6263.

Remaining project failures (24): `#3688`, `#4433`, `#5136`, `#5680`, `#6094`, `_failed_#4569`, `_failed_#5071`, `_failed_cssModuleDuplicate`, `_failed_directives`, `_failed_fragment_props`, `components`, `defineModel`, `directives`, `dynamic-component`, `events`, `fallthroughAttributes`, `fallthroughAttributes_checkRequired`, `fallthroughAttributes_generic`, `script_src`, `slots`, `templateRef`, `unwrapBindingAccess`, `useTemplateRef-dynamic-arg`, `v-generic`.

Next contract repairs include template-ref arrays/unions/generic instances and the flagged `$refs` surfaces. CSS declaration imports (`#5136`) require `resolveStyleImports` plus authored import-path ownership; typing the generated import without mapping its diagnostics/navigation is insufficient. Unknown-component fallback must distinguish unresolved/error types from ordinary `any`; a TypeScript 6.0.3 minimal probe confirms these produce different exact-type diagnostics, but the four `Self` assertions still need an end-to-end reference comparison.

Unknown script languages still need a common whole-SFC checkability contract. JavaScript syntax ownership still uses the parser fallback because the existing checkJs gate suppresses native diagnostics; separating syntactic and semantic gating remains required. Malformed TypeScript now reports native codes, resolving `_failed_#3632` and `_failed_#5823`; `_failed_#5071` still needs exact authored-position repair. An expanded Canon test-wide Clippy run also exposed existing disallowed-macro/type findings beyond the current CI lint scope. NodeNext extensionless JSX imports currently produce TS2834 rather than vue-tsc's TS2835 suggestion. Full patterned-template reference coverage remains #6176; Content Mapper directive ownership remains #6258; rich authoring remains #3953. Keep this issue open until all contracts and the complete CI suite pass.

ubugeeei:
Shared imports could silently become `any`, while editor consumers generated different code and source maps under the same native document identity. Dependency edits then discarded still-open project roots, causing lost diagnostics, stale navigation, repeated project reconstruction and excessive native-process memory.

CLI and editor consumers now retain explicit dependency contracts and a single materialized editor projection. Failed resolution probes invalidate when files appear; closed-file references synchronize the workspace in one batch. Live sources are regenerated together from current buffers before stale contexts are pruned, preserving document identities while real close/delete operations remove obsolete roots. Canon disables automatic type acquisition in its private native project, using the workspace's installed types instead of background npm downloads. Diagnostic queues coalesce completed current-version work without dropping parser feedback or newer edits. Native editor requests use the existing executor to avoid lost wakes after saves (#6263).

Public SFC types preserve actual generic props, emits, expose and inferred/declared slots. Model macros use the installed Vue package’s native overloads instead of handwritten copies. Default-inferred public model props and update events retain their authored setup scope and generic parameters. Maestro maps diagnostics only through explicit Canon source ranges; inference-only probes no longer produce duplicate diagnostics through guessed nearby lines. The obsolete fallback and its redundant SFC parsing are removed. Checks retain callback contextual typing, lexical slot and loop scopes, root-element types, recursive self components and imported-name precedence. Explicit `defineOptions` names resolve recursive props and slots from the recognized macro AST. Runtime builtins, generic directive hooks, reserved Options API members, CSS Modules, aliased Art dependencies and top-level/inherited configuration use the same contracts. Authored diagnostic and rename mappings are exact, including quoted prop keys and shadowed loop sources under patterned templates. Shared compiler analysis preserves slot default/computed-key reads and avoids redundant script/template parsing.

JSX source is lowered once by Canon for CLI and editor requests; Maestro consumes that exact generated code, native identity and source map. Registered module edges are resolved from the authored projection after the complete graph is known. Relative/runtime-extension/alias/barrel/directory/absolute imports retain native extension priority across unsaved edits, file creation/deletion and close-to-disk restoration. Generated JSX remains a plain `.ts` document.

Hover, completion and signature-help requests negotiate Markdown with the native engine. Resolved completions include a syntax-highlightable TypeScript signature alongside authored JSDoc and fenced examples. JSX completions retain lazy resolution data and map a cursor at the end of a member expression correctly. Incomplete SFC scripts keep their authored code and mappings instead of collapsing to an empty component. Analysis-only JSX recovery remaps every temporary parse span to the original source and retains the original syntax errors; compilation still rejects invalid source. Real edit sequences cover template expressions, an incomplete neighboring script, SFC TSX and standalone TSX. The packaged VS Code host now checks readable Markdown, hover ranges, exact jump targets and completion/repair through its actual providers.

The complete pinned Vue Language Tools diagnostic corpus is imported with its MIT license, upstream revision and SHA-256 inventory: **230 projects, 599 original files, 39 expected diagnostics and 71 suppression annotations**. Every project uses its authored configuration and exact diagnostic identities. Separate vue-tsc oracles and mutation/repair tests reject accidental `any`, missing declarations and invalid public contracts. No failing project is excluded or assigned a passing baseline.

Validation for the latest changes:
- Canon library: **1,419 passed**; Croquis: **376 passed**. Native JSX resolution and editor suites: **25 passed**; the final canonical mapping check: **4 passed**. Explicit/implicit/generic recursive-component CLI/live LSP/vue-tsc contracts: **4 passed**.
- A persistent native-session test verifies TSX dependency edits, higher-priority `.ts` creation/removal and repair without restarting the native engine. CLI mixed-project integration and Maestro canonical mapping/JSX regressions pass.
- Actions at `96ac56925`: **68 successful, 3 failed, 8 skipped** checks. Both Criterion A/B and the unchanged PR performance budget pass. Linux full Misskey churn, exact native Content Mapper, app-readiness and real editor host pass. The strict upstream suite and two JSX regressions caused the remaining test/aggregate failures; the JSX repairs are included in the new head.
- Local full Misskey churn passes unchanged 384 MiB/three-process gates: **293.6 MiB** process-tree peak, 40 edit cycles, 290 diagnostic publications and 20 file lifecycle iterations, with default GC and no periodic restart. Median cycle: 382.8 ms.
- Canon/Maestro Clippy with warnings denied, source-length, assertion-lint, formatting and whitespace gates pass. Prior checks also cover the SFC compiler (611), native Art (60), exact references and VS Code host requests. Pinned Misskey check results remove 127 false diagnostics without adding diagnostics.
- Head `7c4e4d5c6` completed both Criterion A/B and PR performance budgets, Content Mapper and app E2E successfully. The remaining non-corpus failure was a stale JSX configuration test assertion, fixed in this head.
- Native model/diagnostic ownership validation: **20 CLI/live LSP/vue-tsc contracts**, **982 Maestro library tests** and **5 assertion lint tests** pass. Both installed Vue versions, generic/defaulted models, transformed reads/writes, declaration emit, authored errors and repair are covered. Canon/Maestro library Clippy with warnings denied and source-length/format gates pass.
- Head `98942f0ca` completed **64 successful, 6 failed and 8 skipped** checks, including both performance gates and Content Mapper. The extra failures were stale generated-model/declaration snapshots; these are updated after verifying authored diagnostics, public component declarations and output files stayed unchanged. The strict upstream failures remain open.
- Rich editor changes pass **22 real LSP authoring/hover/completion/navigation/signature tests**, plus the added incomplete-neighbor case. Canon **1,419**, Maestro **982** and JSX **94** library tests pass. Declaration build integration: **4 passed**. Library Clippy, source lengths, assertion lint and formatting pass. Packaged-host validation is pending Actions.
- Current pushed head: `ffc65975e`; fresh Actions validation is in progress.

**Draft / incomplete:** the strict upstream suite now passes **201/230 projects**; all **29 failures remain enforced in CI**. The remaining component/slot/ref/attrs/event/script-source contracts are tracked in #6253. Native NodeNext extensionless JSX imports are correctly rejected, but Vize reports TS2834 where vue-tsc offers TS2835 with an extension suggestion; the independent regression explicitly records this remaining diagnostic difference. Live-project memory repair #6264 passes Linux Actions, while complete PR validation and merge remain pending. This PR does not close #6253 or alpha epic #6239. Full patterned-template coverage remains #6176; Content Mapper directive ownership remains #6258. Validated Davinci lexical-scope work from #6262 is integrated without claiming overall Davinci completion.

BREAKING CHANGE: public generic SFCs are callable instead of artificially constructible; public emit and slot types now follow Vue semantics; a typed component instance is no longer artificially assignable to an unconstrained `ComponentPublicInstance`. Exhaustive `TemplateExpressionKind` matches must handle `DynamicEventArgument` (existing discriminants are preserved); exhaustive `VizeSemanticLinkKind` matches must handle `VueSetupImportSpecialization`; `CorsaVueVirtualDocumentOptions` adds `jsx_typecheck`; `CorsaScriptVirtualDocument` adds canonical `mappings`; constructing `CorsaVueVirtualDocument` and `CorsaScriptVirtualDocument` requires `resolved_dependencies`; `open_script_virtual_document_with_vue_dependencies` returns `CorsaScriptVirtualDocument` containing the exact native request URI, code and import map instead of only code.


ubugeeei:
Pushed `b178fd598` (on top of the pending `aef2e513f` snapshot fix, which turns `clippy-and-test` green): **the corpus is now 215/230**, up from 206. Every remaining failure stays enforced.

Newly passing: `#5680`, `fallthroughAttributes`, `fallthroughAttributes_checkRequired`, `#4433`, `#6094`, `events`, `_failed_fragment_props`, `script_src`, `#3688`.

| Area | Change |
| --- | --- |
| Fallthrough attributes | component roots resolve to the rendered component's props (`__VizeComponentFallthroughProps`); listener args consult `__vizeFallthroughProps`; `inheritAttrs: false` from the regular `<script>` survives the croquis merge; `checkRequiredFallthroughAttributes` plumbed (declared-minus-authored forwarding, relaxed root usage, `__VizeRequiredFallthroughProps` on parents) |
| Inline component handlers | declared against the listener prop type with its declared return type; lone expression returned, early-exit `v-if` guard; handlers that never read `$event` no longer skip listener typing |
| Duplicate listeners | `@foo-bar` + `@fooBar` (same modifiers) reported as duplicate keys |
| Fragment props | `<template v-if :foo>` checked against the `<template>` element (TS2353) |
| Scripts | non-JS/TS `lang` blocks ignored; `<script src>` imported as the default export |
| Styles | `__VLS_StyleScopedClasses` alias for templates that name it (`resolveStyleClassNames`) |

Still failing (15): `#5136`, `_failed_#4569`, `_failed_#5071`, `_failed_cssModuleDuplicate`, `_failed_directives`, `components`, `defineModel`, `directives`, `dynamic-component`, `fallthroughAttributes_generic`, `slots`, `templateRef`, `useTemplateRef-dynamic-arg`, `unwrapBindingAccess`, `v-generic`.

Local validation: canon 1425 / croquis 381 unit tests, clippy (canon/croquis/maestro, `-D warnings`), rustfmt (style_edition 2024), source-length and assertion gates, the affected `tests/tooling` suites (51) and the CLI content-mapper/check integration tests pass. `event_scope.rs` and `component_prop_checker.rs` were split for the 350-line budget.

ubugeeei:
Pushed `a5572a518`: **corpus 217/230** (adds `dynamic-component` and `useTemplateRef-dynamic-arg`), and the three real-project regressions the previous head surfaced in Actions are fixed and reproduced locally with the freshly built binary.

| Actions finding | Cause | Fix |
| --- | --- | --- |
| TS1117 duplicate listeners (misskey / reka-ui / nuxt-ui) | `v-model` records an `update:` listener like an authored one, so `v-model` + `@update:model-value` collided | v-model-derived listeners never join the duplicate-key check |
| TS7006 on `(event) => …` (reka-ui) | a zero-parameter forwarded listener imposed a zero-argument contextual type | such a signature leaves the listener untyped |
| five missing-required-prop errors lost (nuxt-ui) | a root whose module does not resolve makes `typeof Root` the error type, which conditionals and homomorphic mapped types pass through as `any` | the forwarded surface is a non-homomorphic mapped type, degrading to an index signature instead |

A fourth one showed up only when checking reka-ui from the monorepo root (no tsconfig, alias imports unresolved): a grandchild's `PointerEvent` payload won over an unresolved declared emit. Component-root resolution and fallthrough listener typing are now applied only under `fallthroughAttributes`, the option `vue-tsc` requires for the same inference; native-root fallthrough props keep their previous unconditional shape.

Also in this push: `<component :is="cond ? Foo : Bar">` is aliased at the template root so its props, listeners and slots are checked against the union; a dynamic argument's TS2464 anchors at the directive prefix like `vue-tsc`; `__VLS_StyleScopedClasses` is a module-level alias; `ResolveStyleClassNames` is `native`-gated (the workspace Clippy build compiles without that feature — the reason `clippy-and-test` failed); three files split for the 350-line budget.

Local validation at `a5572a518`: canon 1427 / croquis 381 / maestro davinci_ts40 3 unit tests, `cargo clippy --workspace -D warnings` (the CI invocation), rustfmt style_edition 2024, source-length gate, the reka-ui check snapshot against the debug binary (`VIZE_BIN`), `tests/snapshots/check/ecosystem-products` regenerated with identical diagnostics. Still failing (13): `#5136`, `_failed_#4569`, `_failed_#5071`, `_failed_cssModuleDuplicate`, `_failed_directives`, `components`, `defineModel`, `directives`, `fallthroughAttributes_generic`, `slots`, `templateRef`, `unwrapBindingAccess`, `v-generic`.

ubugeeei:
## Corpus progress: 222 / 230 (head `1723e30f1`)

Two more batches since the last note (`e13da02eb`, `1723e30f1`), all still enforced with no allowlist.

| Project | Cause | Fix |
| --- | --- | --- |
| `#5136` | a CSS module's `@import` made `$style` an open record | `resolveStyleImports`: imported stylesheets contribute their default export's class names |
| `_failed_cssModuleDuplicate` | duplicate `<style module>` names were silently merged | one declaration per block, mapped to the `module` attribute, so the checker itself reports TS2300 / TS2717 |
| `unwrapBindingAccess` | a `v-for` destructuring default (`{ val = fallback }`) read the raw `Ref`; a `Ref \| undefined` binding kept `undefined` where `vue-tsc` reads it through `.value` | pattern defaults are template reads; maybe-ref bindings read in a narrowing position use the `.value` semantics (nullish absorbed only when every other member is a ref) |
| `v-generic` | `<!-- @vue-generic {T} -->` was ignored | the usage resolves through the component instantiated with the authored arguments; constraint errors report at the comment and the usage's `@vue-expect-error` owns it |
| `slots` | an unregistered tag was a declared `any`, so `exactType(slotProp, …)` failed a `never` parameter | unresolved tags are error types (`{}["Tag"]` under `@ts-ignore`), which TypeScript never reports on - verified against real `vue-tsc` 3.3.11 on the same fixture |

Also repaired baselines CI had not reached behind earlier fail-fast failures: `build_declaration_emit_cli` / `check_can_emit_declarations` / patterned root snapshots, the TS-40 legacy projection, the Croquis folio fixtures, the generated Davinci inventories, and the source length budget for `nuxt_false_positives.rs`.

### Remaining 8

- `components`, `defineModel`: `exactType(ScriptSetup, defineComponent({...}))` needs the SFC default export to be **type-identical** to Vue's `DefineComponent`. Vize's `__vize*` marker members on the export break that identity by construction. Closing these means moving the markers off the exported value type (a merged namespace or a side channel) and deriving the export from `defineComponent` - an export-shape redesign with type-check performance implications, so it deserves a maintainer decision before I start.
- `templateRef`: `v-for` ref arrays, duplicate-name unions, unknown-key errors for top-level `useTemplateRef` calls only, generic instance exposure, `$refs` inference options.
- `directives`: arg / modifier typing, `GlobalDirectives`, Options API `directives`, the Vapor call form, and `typeof X` inside a hoisted `declare module`.
- `fallthroughAttributes_generic`: generic type propagation through a component fallthrough root.
- `_failed_directives`: unknown template names must be TS2339 (property on the instance) instead of TS2304 in `<script setup>` too - the Options API path already does this (#3888); broad snapshot churn.
- `_failed_#4569`, `_failed_#5071`: upstream layout artifacts (an empty `generic` attribute; a hoisted import glued before `<script>` content).

